PR #5, opened 2016-09-02
Close widget when figure is closed from python

jenshnielsen:
This ensures that the figure in the notebook is closed correctly when closed 

I.e. 
```
plt.plot(range(10))
```
creates a figure and then in a new cell. Assuming this is the first figure we can close it whit

```
plt.close(1)
```
In a new cell. Currently this does not have any visual effect. This change ensures that the figure is closed in the notebook.

I have removed the `self._send_event('close')` because it seems to trigger a no comms warning. As I understand it this no longer needed anyway as the widget handles the closing.


SylvainCorlay:
This looks good to me.

PR #6, opened 2016-10-06
API: change focus policy + rm nb special case

tacaswell:
Do not focus on creation, user must actively click on the figure
to get focus (and hence have the figure keyboard events).

This allows us to remove the shift-enter special case (which will be
configurable in jh).

Disable propagation on any keyboard events that the figure gets, which
removes the need for the keyboard_manager special casing.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -10,7 +10,8 @@
     "url": "https://github.com/matplotlib/jupyter-matplotlib.git"
   },
   "scripts": {
-    "prepublish": "webpack",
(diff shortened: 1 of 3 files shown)


jasongrout:
Thanks!

jasongrout:
oh...I can't merge. I can't test just yet, but the code looks good to me.

jasongrout:
@tacaswell - can you merge this, and then I'll rebase the Lab work on master? (I still can't merge it :).

Issue #9, opened 2017-02-10
notebook backend not working in jupyterlab


michaelaye:
I just installed this on top of my standard env, using conda-forge.

Using standard `%matplotlib nbagg` as activation command and using standard mpl plotting did not pop up any figure inside a jupyterlab notebook?

Using MPL 2.0 and jupyterlab v0.16 on a Python 3.5 env with conda on OSX 10.11.6 inside Safari 10.0.3

SylvainCorlay:
All you need to enable this is to import ipympl.

michaelaye:
Apparently, it must be done at the right time?

```python
/Users/klay6683/miniconda3/envs/stable/lib/python3.5/site-packages/matplotlib/__init__.py:1401: UserWarning:  This call to matplotlib.use() has no effect
because the backend has already been chosen;
matplotlib.use() must be called *before* pylab, matplotlib.pyplot,
or matplotlib.backends is imported for the first time.

  warnings.warn(_use_error_msg)
```

michaelaye:
Still doesn't work. I put `import ipympl` now as the first line of import in my .ipython profile startup, but still no plot is popping up in jupyterlab v0.16.

SylvainCorlay:
@michaelaye I am sorry I replied from a mobile device earlier and missed that you were using jupyterlab.

We will add support for jupyterlab after we release ipywidgets 6.0.

thomasaarholt:
Could someone edit the title to "Notebook backend not working"? 
It would be nice to get the notebook/nbagg backend working in jupyter lab. I know many users enjoy having an inline interactive backend.

danielballan:
Confirmed that it is still broken in the latest release, 0.21.0. Will try to look at this with @tacaswell next week.

ElmWer:
still not working (ipywidgets 6.0.0, matplotlib  2.0.2, jupyter lab: 0.25.2, Python 3.5.3). It would be appreciated to use nbagg in jupyter lab.

tacaswell:
Hopefully this will get worked on at the jupytcon sprints next month.

madx:
Hi, any news about this issue?

blink1073:
This is now working (shown with pinned versions so this will always work):

```
pip install jupyterlab==0.28.11 ipympl==0.0.8
jupyter labextension install @jupyter-widgets/jupyterlab-manager@^0.28 jupyter-matplotlib@^0.1
jupyter lab
```

```python
[...]

PR #21, opened 2017-10-28
Update for ipywidgets 7.0 and JupyterLab 0.28

### LINKED ISSUE #15: compatibility with ipywidgets-7.0.0 and later ?
With latest ipywidgets-7.0.0b6 , the example https://github.com/matplotlib/jupyter-matplotlib/blob/master/examples/ipympl.ipynb runs srtangely on classic notebook (not really usable).

blink1073:
Fixes #15. 

- Removed export button and fixed the handling of interactivity toggling.
- Fixed the download image behavior since Chrome no longer supports data urls for `window.open`.

mpl-widget

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Matplotlib Jupyter Extension.
 This repository contains code for the Matplotlib Jupyter widget, stripped out
 of the main matplotlib repository.
 
-It requires matplotlib 2.0.0 or more recent.
+It requires matplotlib 2.0 or and ipywidgets 7.0 more recent.
 
 The goal of this project is to separate developement of the Jupyter integration
 (future versions of notebook and Jupyter Lab) from the calendar of the releases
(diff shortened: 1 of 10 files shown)


blink1073:
@tacaswell @SylvainCorlay, I had one bug that I can't figure out.  I sometimes get the following error that is cleared by restarting the kernel:

image

blink1073:
cc @jasongrout

blink1073:
Note that the video above shows it creating a new window to download.  I changed it so that it automatically downloads '<figure_title>.png` without navigating away from the page.

blink1073:
Note that the following also works (should this close #9)?:

```python
%matplotlib ipympl

plt.plot([0, 1, 2, 2])
plt.show()
```

blink1073:
Found the source of the usage error:

https://github.com/ipython/ipython/blob/dfb458636b247a78f4a57206128caffd97035ea5/IPython/core/pylabtools.py#L27

Ensured that we always get "ipympl" in 676ea16.

blink1073:
Ah, the problem is that the addition of `widget` in `IPython.core.pylabtools` needed a corresponding addition to https://github.com/ipython/ipykernel/blob/71c44fe80e79126de4f27e6713fda1d274132d77/ipykernel/eventloops.py#L45.

blink1073:
PR: https://github.com/ipython/ipykernel/pull/275

SylvainCorlay:
Goog catch for the IPython backend name!

blink1073:
Sweet!  Can you add me to the PyPI maintainers?

SylvainCorlay:
Done!

PR #35, opened 2017-12-15
Support the matplotlib widget backend

blink1073:
Thanks to the `ipykernel@4.7` [release](https://github.com/ipython/ipykernel/pull/275), we can now use:

```python
%matplotlib widget

plt.plot([0, 1, 2, 2])
plt.show()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ of the main matplotlib repository.
 
 It requires matplotlib 2.0 or and ipywidgets 7.0 more recent.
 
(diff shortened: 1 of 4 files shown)


stonebig:
remark: if you try to update the top cell in this example, this is the bottom cell that is re-drawn.
ipympl_bb

tacaswell:
@stonebig That may be the correct behavior depending on what the 'current' axes is.

SylvainCorlay:
Closing as rebased in #54.

PR #82, opened 2019-06-19
Use ipywidgets CSS for controls

martinRenou:
Use ipywidgets CSS for controls, for a modernized style.

test_ipympl


SylvainCorlay:
Thanks, this looks great!

How about adding the same CSS as the Label widget for the text area in the upper right?

This would align the text vertically with the buttons and ajust the line height.

SylvainCorlay:
Merging as a net improvement.

martinRenou:
> How about adding the same CSS as the Label widget for the text area in the upper right?

Will do

PR #84, opened 2019-06-20
Cleaning

martinRenou:
Just cleaning code:
- Remove unused functions
- Move everything concerning the figure prototype inside of `mpl.js`, now `mpl_widget.js` only contains the `MPLCanvas` classes
- Some functions were defined in `mpl.js`, and overwritten in `mpl_widget.js`, I simply replaced the implementation from `mpl.js` by the ones in `mpl_widget.js`

### DIFF
--- a/js/src/mpl.js
+++ b/js/src/mpl.js
@@ -14,19 +14,6 @@
 window.mpl = {};
 
 
-mpl.get_websocket_type = function() {
-    if (typeof(WebSocket) !== 'undefined') {
-        return WebSocket;
-    } else if (typeof(MozWebSocket) !== 'undefined') {
-        return MozWebSocket;
-    } else {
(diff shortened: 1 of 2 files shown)


js/src/mpl.js
```diff
@@ -90,11 +77,20 @@ mpl.figure.prototype._init_header = function() {
...
+    // this is important to make the div 'focusable'
+    canvas_div.attr('tabindex', 0);
 }
 
 mpl.figure.prototype._root_extra_style = function(canvas_div) {
+    var fig = this;
+    canvas_div.on("remove", function(){
+        fig.close_ws(fig, {});
+    });
+}
```
SylvainCorlay: remove _root_extra_style  and close_ws.

PR #93, opened 2019-06-21
Fix save

martinRenou:
The "diff" image was saved after pressing the `download` button, not the actual plot


tfblum:
The code as written when this file was changed shows the filename always ending with '.png' but shouldn't this change depending on the default value listed in the matplotlibrc file?

thomasaarholt:
This is a good point. The relevant entry is in `plt.rcParams['savefig.format']`, but I'm not sure how that translates to the javascript code.

martinRenou:
Do you mind opening a specific issue for this? :)

thomasaarholt:
Done #138

PR #94, opened 2019-06-21
Remove scrolling logic

### LINKED ISSUE #81: plots capture scrolling
scrolling while the mouse is over a plot doesn't scroll in the document (the notebook activity indicator seems to suggest that "scroll events" are processed by the widget). However nothing happens (no zooming into the plot etc.). 
Not being able to scroll while the mouse is on some parts of the notebook is very annoying for me.

Is it possible to implement an option which would disable this?

martinRenou:
The scrolling does not seem to work anyway. It should close #81


tacaswell:
Also left a comment on #81 , but I suspect that the it was working and the scroll events were making it back to the figure on the python side, but just had no action subscribed to it on the python side.

PR #97, opened 2019-06-24
Remove jupyter.setDirty call

martinRenou:
From what I understand, this is not useful


SylvainCorlay:
We may want to re-add it when adding the image mime type representation of figures.

SylvainCorlay:
Merging as per in-person discussion.

PR #99, opened 2019-06-25
Interactive Toolbar widget

martinRenou:
Make the toolbar an interactive widget.

Now `fig.canvas.toolbar` is a widget, and I also added `fig.canvas.toolbar_visible` and `fig.canvas.toolbar_position` for allowing to hide/move the Toolbar dynamically

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -8,8 +8,11 @@ from uuid import uuid4 as uuid
 
 from IPython.display import display, HTML
 
-from ipywidgets import DOMWidget
-from traitlets import Unicode, Bool, Float, List, Any
+from ipywidgets import DOMWidget, widget_serialization
+from traitlets import (
+    Unicode, Bool, Float, List, Any, Instance,
+    default
+)
 
 from matplotlib import rcParams
 from matplotlib.figure import Figure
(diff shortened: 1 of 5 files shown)


ipympl/backend_nbagg.py
```diff
@@ -21,6 +24,10 @@
 from matplotlib.backend_bases import (ShowBase, NavigationToolbar2,
                                       FigureCanvasBase)
 
+here = os.path.dirname(__file__)
+with open(os.path.join(here, 'static', 'package.json')) as fid:
+    js_version = '^%s' % json.load(fid)['version']
```
SylvainCorlay: quick nitpick: maybe `js_semver` instead of `js_version`.

SylvainCorlay: (since it is not a version but a semver range)

js/src/mpl.js
```diff
@@ -253,7 +189,21 @@ figure.prototype.handle_rubberband = function(msg) {
 
 figure.prototype.handle_figure_label = function(msg) {
     // Updates the figure title.
-    this.header.textContent = msg['label'];
+    this.figure_label = msg['label'];
+    this.update_header();
```
SylvainCorlay: I think the message info should be in a separate status bar rather than adding text to the title.

martinRenou: It looks better like that indeed

martinRenou:
Thanks! :)

PR #100, opened 2019-06-27
Figure widget

martinRenou:
This moves the `figure` class implementation to the `Canvas` widget


js/src/mpl.js
```diff
@@ -1,8 +1,12 @@
+var _ = require('lodash');
+var Backbone = require('backbone');
 var utils = require('./utils.js');
 
-figure = function(figure_id, widget) {
+figure = function(figure_id) {
+    // Make the figure an eventful object
+    _.extend(this, Backbone.Events);
```
martinRenou: This should be removed when the `figure` will actually be an interactive widget

PR #104, opened 2019-06-28
Center toolbar buttons

martinRenou:


SylvainCorlay:
Note: I am actually the person who pressed the green button from @martinRenou's laptop.

PR #105, opened 2019-06-28
Furthering examples with a quick and raw plot and changing output size

fdobad:
Hello everyone, I hope everyone has a great day when reading this.

Using jupyterlab, I wanted a quick way to plot a set of data with a shared axis, but disjoint values by putting each extra series with its own axis, avoiding scaling issues. (For example if one series is thousands of dollars, and the other in [0,1], you won't see them right), this image explains it better:
Capture2

So you can see I used some ipywidgets components to interact, so I think this is a good contribution to the examples.

At the bottom of the notebook I left some questions, that (being my first contribution) thought would be the best place to leave them for discussion, summarizing:

    1. There's a slightly annoying bug that displays 2 graphs instead of one, before interacting

    2. The newly created axes don't update on zoom as the normal one (on the left) does

    3. Is there a way stretch out the graph (canvas?) to the output size and have a bottom right stretch control to resize?


martinRenou:
Your example looks good!! I just wanted to answer the last question, we are working on the resize of the plot currently, so it _should_ be possible soon. Though I'm not sure it is part of the plan to add a stretch control, but you should be able to control the size through the [widget layout](https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Styling.html).

fdobad:
Hi! great news that you're working on the resize through widget layout. Now I can stop trying! lol! Although in all seriousness thanks for your work on this packages... I'm really grateful and still excited how jupyter(lab,ipywidgets) and matplotlib has made paperthin all ui-related software for us data scientists and engineers.

As for the example, do that 'looking good' of it mean that hopefully someone will slash my commit to add it on the next release? Or is there more stuff I should do to this process?

Cheers!

martinRenou:
I'm happy to work on this package! :)

> There's a slightly annoying bug that displays 2 graphs instead of one, before interacting

I saw this bug. Did not take time to look at it yet. Would you mind opening an issue for that if that's not the case already?

Concerning your commit, would you mind removing the `TODO help needed` section from the Notebook? I don't think it belongs in the example Notebook of the repo. Also, maybe you could make it a separate Notebook? Something like `advanced.ipynb`? Because your example is a bit more code than the other ones before getting a plot displayed on the screen.

fdobad:
> I saw this bug. Did not take time to look at it yet. Would you mind opening an issue for that if that's not the case already?

I gave it another thought, from the ipywidgets docs read: "interact autogenerates UI controls" and "interactive returns a Widget instance", so I tried using interactive instead, but didn't redraw the plot, collapsing its space instead.

This is the change: From display to return on the call function, and the interact decorator changes to interactive at the bottom.
`    return VBox([plt.figure(1).canvas,HTML(paranoia)])
    
interactive_plot = interactive(on_value_change, w= IntSlider(description='x range',min=2,max=100,step=1,value=4,continuous_update=False),
           x= FloatSlider(description='subplots_adjust(right',min=0.1,max=1,step=0.02,value=0.60,continuous_update=True,style = {'description_width': 'initial'}),
           y= FloatSlider(description='offset',min=15,max=60,step=2,value=35,continuous_update=False),
           z= IntSlider(description='number of lines',min=2,max=10,step=1,value=6,continuous_update=False,style = {'description_width': 'initial'}))
display(interactive_plot )`

So I guess interact works better than interactive for now. I haven't looked into the observe method yet.

Lastly I ran the interact code in the console, showing: that the rogue plot is at beginning, pic:
Capture3.

So I'm not sure if it's an issue or just me using the wrong tool.

PR #109, opened 2019-07-01
Put back the scrolling support

martinRenou:
cc. @tacaswell 

As discussed here https://github.com/matplotlib/jupyter-matplotlib/issues/81#issuecomment-506748493


martinRenou:
Tested with this example: https://gist.github.com/tacaswell/3144287 and it seems to work just fine :)

SylvainCorlay:
@tacaswell what is the way you would like to see this go?

Should the ipympl widget always catch scroll events and we only process them in the backend if the user does something like in your example?

tacaswell:
Yes, this is consistent with how the rest of the interactive backends handle scroll events.

Issue #134, opened 2019-08-22
Change/disable header


maartenbreddels:
The Figure ID is shown in the header (Figure 14):
![image](https://user-images.githubusercontent.com/1765949/63501218-8e850c00-c4cb-11e9-82f7-0f91d576ed9c.png)

Is there an option to change or remove/hide it?

martinRenou:
No, but that could be a nice addition

freethebee:
I did a quick and dirty fix, but still need to check if I didn't miss anything, write tests and see how to do a proper pull request.

```
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -149,6 +149,9 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
     toolbar = Instance(Toolbar, allow_none=True).tag(sync=True, **widget_serialization)
     toolbar_visible = Bool(True).tag(sync=True)
     toolbar_position = Enum(['top', 'bottom', 'left', 'right'], default_value='left').tag(sync=True)
+
+    header_visible = Bool(True).tag(sync=True)
+
     _closed = Bool(True)

     # Must declare the superclass private members.
diff --git a/js/src/mpl_widget.js b/js/src/mpl_widget.js
index c8c8585..ee547cd 100644
--- a/js/src/mpl_widget.js
+++ b/js/src/mpl_widget.js
@@ -16,6 +16,7 @@ var MPLCanvasModel = widgets.DOMWidgetModel.extend({
             _view_module: 'jupyter-matplotlib',
             _model_module_version: '^'+ version,
             _view_module_version: '^' + version,
+            header_visible: true,
             toolbar: null,
             toolbar_visible: true,
             toolbar_position: 'horizontal'
@@ -64,6 +65,7 @@ var MPLCanvasView = widgets.DOMWidgetView.extend({

     model_events: function() {
         this.model.on('msg:custom', this.on_comm_message.bind(this));
+        this.model.on('change:header_visible', this.update_header_visible.bind(this));
         this.model.on('change:toolbar_visible', this.update_toolbar_visible.bind(this));
         this.model.on('change:toolbar_position', this.update_toolbar_position.bind(this));
     },
@@ -78,6 +80,11 @@ var MPLCanvasView = widgets.DOMWidgetView.extend({
         this.send_message('initialized');
     },

+    update_header_visible: function() {
+        this.header.style.display = this.model.get('header_visible') ? '': 'none';
[...]

Issue #127, opened 2019-07-16
RuntimeError: libpng signaled error


chaffra:
Sometimes I am getting this error with the trunk version. Not use what a MWE would be. Has anyone encountered this?
Thanks, Chaffra
```
RuntimeError                              Traceback (most recent call last)
/usr/local/lib/python3.6/dist-packages/ipympl/backend_nbagg.py in _handle_message(self, object, content, buffers)
    178             self.manager.resize(w, h)
    179         else:
--> 180             self.manager.handle_json(content)
    181 
    182     def send_json(self, content):

/usr/local/lib/python3.6/dist-packages/matplotlib/backends/backend_webagg_core.py in handle_json(self, content)
    437 
    438     def handle_json(self, content):
--> 439         self.canvas.handle_event(content)
    440 
    441     def refresh_all(self):

/usr/local/lib/python3.6/dist-packages/matplotlib/backends/backend_webagg_core.py in handle_event(self, event)
    238         handler = getattr(self, 'handle_{0}'.format(e_type),
    239                           self.handle_unknown_event)
--> 240         return handler(event)
    241 
    242     def handle_unknown_event(self, event):

/usr/local/lib/python3.6/dist-packages/matplotlib/backends/backend_webagg_core.py in handle_draw(self, event)
    254 
    255     def handle_draw(self, event):
--> 256         self.draw()
    257 
    258     def _handle_mouse(self, event):

/usr/local/lib/python3.6/dist-packages/matplotlib/backends/backend_webagg_core.py in draw(self)
    151             super().draw()
    152         finally:
--> 153             self.manager.refresh_all()  # Swap the frames.
    154 
    155     def draw_idle(self):

/usr/local/lib/python3.6/dist-packages/matplotlib/backends/backend_webagg_core.py in refresh_all(self)
    441     def refresh_all(self):
    442         if self.web_sockets:
--> 443             diff = self.canvas.get_diff_image()
    444             if diff is not None:
    445                 for s in self.web_sockets:

/usr/local/lib/python3.6/dist-packages/matplotlib/backends/backend_webagg_core.py in get_diff_image(self)
    199             buff = _png.write_png(
    200                 output.view(dtype=np.uint8).reshape(output.shape + (4,)),
--> 201                 None, compression=6, filter=_png.PNG_FILTER_NONE)
    202 
    203             # Swap the renderer frames

RuntimeError: libpng signaled error
```

thomasaarholt:
[...]

PR #135, opened 2019-08-23
Hide header

freethebee:
This branch contains a fix for hiding the figure header in the matplotlib widget, as discussed in issue #134 . I also added an example notebook to show how it works.
I have some issues with libpng errors, but I am not sure how they correspond to my code, as I see those in other cases as well. I will explain in the issue comments.

### DIFF
--- a/js/src/mpl_widget.js
+++ b/js/src/mpl_widget.js
@@ -16,6 +16,7 @@ var MPLCanvasModel = widgets.DOMWidgetModel.extend({
             _view_module: 'jupyter-matplotlib',
             _model_module_version: '^'+ version,
             _view_module_version: '^' + version,
+            header_visible: true,
             toolbar: null,
             toolbar_visible: true,
             toolbar_position: 'horizontal'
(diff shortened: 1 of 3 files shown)


js/src/mpl_widget.js
```diff
@@ -78,6 +80,11 @@ var MPLCanvasView = widgets.DOMWidgetView.extend({
         this.send_message('initialized');
     },
 
+    update_header_visible: function() {
+        this.header.style.display = this.model.get('header_visible') ? '': 'none';
```
martinRenou: You might want to do this when creating `this.header` as well :) In case `header_visible` is False when creating the canvas.

freethebee: Conda was angry at me and it took some effort to be up and running again.

I can add the call to the _init_header method, but the request_resize() causes a libpng error if header_visible is set to False before rendering (#127). 
I tried adding,
```
this.header.style.display = this.model.get('header_visible') ? '': 'none';
```
to _init_header(), while leaving the update method as it is, but this does not change anything. 
I find it a bit strange as the events seem to be set after _init_header(), but unfortunately my javascript knowledge is limited.
The same problem exists for hiding the toolbar, so solving this one would improve that option as well. If you have any suggestions?

martinRenou: I would say let's merge it as it is for now. It looks good to me, we'll see later if my comment is needed.

martinRenou:
Thanks again for your PR!

PR #152, opened 2019-11-16
no attribute '_send_event' problem in a blind way

stonebig:
at least now, there is no more https://github.com/matplotlib/jupyter-matplotlib/issues/115:
- I have my figures,
- It didn't remove last "improvement"

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -212,7 +212,7 @@ class FigureManager(FigureManagerWebAgg):
     def destroy(self):
         try:
             self.canvas.close()
-        except:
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
@stonebig I lack a bit of context here. Could you ellaborate a bit on the logic?

SylvainCorlay:
(And sorry for the late reply, we were all swamped with work the past few weeks).

stonebig:
hum, with latest ipympl-04., I get the error if I run all my winpython-checker.ipynb,  don't get it if I run individual cells. droping my fix, as it may not be the correct problem

Issue #164, opened 2020-01-17
Outdated stable branch and malfunctional mybinder.org badge


consideRatio:
Visiting this repos mybinder example, both using the old stable branch as in the README, or the current master branch, gives us:

![image](https://user-images.githubusercontent.com/3837114/72602142-12c2aa80-3917-11ea-9403-930f15dfc009.png)

![image](https://user-images.githubusercontent.com/3837114/72602528-b57b2900-3917-11ea-8075-ba24e9d8f2d3.png)

This is because mybinder.org won't automatically realize it should install the jupyterlab extension. We would need an apt.txt file in the repo for that to happen where we write:

```
jupyter labextension install @jupyter-widgets/jupyterlab-manager
jupyter labextension install jupyter-matplotlib
```

Also, update the readme to use the master branch may be simpler than remembering to update the stable branch?

martinRenou:
Thanks for the report. Would you mind opening a PR for the jupyterlab build?

The reason for the stable branch is that if the master branch moves too much, users don't need to rebuild the docker image on binder

PR #165, opened 2020-01-17
Fix binder badge, add travis/conda/npm badges

consideRatio:
Addresses #164, but the stable branch is still outdated so if anyone builds from that it is not up to date with 0.4.0 for example.


martinRenou:
Thanks a lot !

PR #59, opened 2018-07-14
Add text-html repr

SylvainCorlay:

PR #176, opened 2020-02-25
Stop button

### LINKED ISSUE #16: Images not saved in *.ipynb
Maybe its a design decision, but if I use `%matplotlib ipympl` in a notebook, there is no image of plots I've made is saved in the *.ipynb file as there was for `nbagg`. Thats a huge usability issue for me.

I think the cool thing about notebooks is being able to see the output of your work without having to re-run all the code, some of which may no longer work (data moved etc), so this is a pretty big reversion in functionality for me.  It also kills the ability to quickly post onto GitHub or gist, and even save as html no longer has a plot.  If this is to replace `nbagg` I hope the image saving gets added back in.  

Of course, maybe I'm just doing something incorrectly.  This was with master on matplotlib and master on ipympl.  Jupyter-notebook = 5.0.0

### LINKED ISSUE #150: embed and close button
It would be really nice to have an 'embed and close' button in the toolbar, which closes the figure canvas, removes the widget and embeds the corresponding png into the notebook. That way, the figures will also persist when the notebook is closed and opened again etc.

martinRenou:
This should fix #150 and fix #16 

It adds a "Stop interactions and embed" button which allows saving the plot as a static image in the Notebook. The output actually stays a widget, which has the last state of the plot in its model.

embed

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -152,7 +152,12 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
 
     header_visible = Bool(True).tag(sync=True)
 
+    width = Float(0, read_only=True).tag(sync=True)
+    height = Float(0, read_only=True).tag(sync=True)
+
     _closed = Bool(True)
+    closed = Bool(False, read_only=True).tag(sync=True)
+    _data_url = Unicode(None, allow_none=True, read_only=True).tag(sync=True)
 
     # Must declare the superclass private members.
     _png_is_old = Bool()
@@ -174,9 +179,7 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
 
     def _handle_message(self, object, content, buffers):
         # Every content has a "type".
-        if content['type'] == 'closing':
-            self._closed = True
-        elif content['type'] == 'initialized':
+        if content['type'] == 'initialized':
             _, _, w, h = self.figure.bbox.bounds
             self.manager.resize(w, h)
         else:
--- a/js/src/mpl_widget.js
+++ b/js/src/mpl_widget.js
@@ -19,6 +19,10 @@ var MPLCanvasModel = widgets.DOMWidgetModel.extend({
             header_visible: true,
             toolbar: null,
             toolbar_visible: true,
+            closed: false,
+            _data_url: null,
+            width: 0,
+            height: 0,
             toolbar_position: 'horizontal'
         });
     }
@@ -38,7 +42,6 @@ var MPLCanvasView = widgets.DOMWidgetView.extend({
         this.image_mode = 'full';
 
         this.figure = document.createElement('div');
-        this.figure.addEventListener('remove', this.close.bind(this));
         this.figure.classList = 'jupyter-matplotlib-figure jupyter-widgets widget-container widget-box widget-vbox';
 
         this._init_header();
@@ -52,6 +55,8 @@ var MPLCanvasView = widgets.DOMWidgetView.extend({
 
         return this.create_child_view(this.model.get('toolbar')).then(function(toolbar_view) {
             that.toolbar_view = toolbar_view;
+
+            toolbar_view.on('close',that.handle_close.bind(that));
 
             that.update_toolbar_position();
 
@@ -72,6 +77,10 @@ var MPLCanvasView = widgets.DOMWidgetView.extend({
     },
 
     send_initialization_message: function() {
+        if (this.model.get('closed')) {
+            return;
+        }
+
         if (this.ratio != 1) {
             this.send_message('set_dpi_ratio', {'dpi_ratio': this.ratio});
         }
@@ -86,7 +95,7 @@ var MPLCanvasView = widgets.DOMWidgetView.extend({
     },
 
     update_toolbar_visible: function() {
-        this.toolbar_view.el.style.display = this.model.get('toolbar_visible') ? '' : 'none';
+        this.toolbar_view.el.style.display = this.model.get('toolbar_visible') && !this.model.get('closed') ? '' : 'none';
     },
 
     update_toolbar_position: function() {
@@ -124,6 +133,14 @@ var MPLCanvasView = widgets.DOMWidgetView.extend({
         while (this.el.firstChild) {
             this.el.removeChild(this.el.firstChild);
         }
+    },
+
+    handle_close: function() {
+        this.model.set('_data_url', this.canvas.toDataURL());
+        this.model.set('closed', true);
+        this.model.save_changes();
+
+        this.update_toolbar_visible();
     },
 
     _init_header: function() {
@@ -208,12 +225,21 @@ var MPLCanvasView = widgets.DOMWidgetView.extend({
                 // almost always do), so we need to clear the canvas so that
                 // there is no ghosting.
                 that.context.clearRect(0, 0, that.canvas.width, that.canvas.height);
+                if (that.model.get('width') != that.canvas.width || that.model.get('height') != that.canvas.height) {
+                    that._resize_canvas(that.model.get('width'), that.model.get('height'));
+                }
             }
             that.context.drawImage(that.image, 0, 0);
         };
 
         this.image.onunload = function() {
             that.close();
+        }
+
+        // Draw saved state if the communication is closed
+        if (this.model.get('closed')) {
+            this.image_mode = 'full';
+            this.image.src = this.model.get('_data_url');
         }
     },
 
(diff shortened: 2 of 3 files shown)


martinRenou:
cc. @tacaswell @SylvainCorlay 

One idea would be to remove the interactive widget `repr` from the mimebundle repr when [the Matplotlib figure is closed](https://matplotlib.org/2.1.0/api/_as_gen/matplotlib.pyplot.close.html), and save the image repr instead.

Not sure I have all the implementation details in mind, but I feel like that would be an elegant way of solving this issue.

xeliba:
This is a really important feature.

After "closing" , all plots should be simple PNG files in the Notebook. That enables us to export the whole notebook as html, pdf etc. Is this workflow already possible with this PR? What is still missing?

(By the way, with the matplotlib 'notebook' backend this is already possible for normal 'Jupyter' notebooks. Not for 'JupyterLab', yet.)

martinRenou:
> Is this workflow already possible with this PR?

This PR is not really fixing this issue properly.
It only saves the image as part of the widget model. The problem is that the widget model is not always saved in the Notebook, so you might not be able to retrieve the image.

> What is still missing?

I think we should take a different approach than what this PR is doing. The approach would be the one I tried to explain on the comment on top of yours:

When the close button is pressed, or when [the Matplotlib figure is closed using the Python API](https://matplotlib.org/2.1.0/api/_as_gen/matplotlib.pyplot.close.html), we should actually remove the Widget and replace it with a plain image in the cell output.

xeliba:
Recently, I tried to obtain the PNG using the renderer of the 'widget' backend obtained with 'fig.canvas.get_renderer()' . But it only returned a buffer with all pixels of the same color, i.e. without the actual plot. Don't know what I have to call in advance. 'fig.canvas.draw()' does not help, which is little bit confusing.

ianhi:
> When the close button is pressed, or when the Matplotlib figure is closed using the Python API, we should actually remove the Widget and replace it with a plain image in the cell output.

Big fan of this feature existing. I'm willing to help - but any advice on what the appropriate way to accomplish this is would be great. Is this PR: https://github.com/matplotlib/ipympl/pull/59 the approach to take?

ianhi:
Looking at it further I do think that #59 shows how to do this. Thinking out loud here: I think that the `Widget` class defines `_repr_mime_bundle` which will normally implant a view of widget into the notebook. So canvas should override this function with some logic that looks like this:
```python
if not self.closed:
    Super()._repr_mimebundle()
else:
    "<img src='data:image/png;base64,{......}'/>"
```

ianhi:
In https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#functions it says that:
> A single object can declare some or all of these representations; all are handled by IPython’s display system.

referring to this: 
image

but if you define multiple of them then it's not clear how to signal which one should take precedence.

martinRenou:
> if you define multiple of them then it's not clear how to signal which one should take precedence

I think that is actually the problem with the PR you pointed to. If I'm correct, it's the UI that chooses which representation to display. I think the widgets representation would always take precedence over the HTML one.

>  I'm willing to help - but any advice on what the appropriate way to accomplish this is would be great. Is this PR: #59 the approach to take?

Honestly, I really don't know. I would love to see a feature like this. And I think it's a big blocker for people to switch from the inline back-end to the widget one.

If you want to implement this you are very welcome to do so! :)

I am wondering if we could use a smart `IPython.display.update_display` call to update the bundle and remove the widget representation when the figure is closed.

ianhi:
Some issues that look relevant:
https://github.com/jupyter-widgets/ipywidgets/issues/2280
https://github.com/jupyter-widgets/ipywidgets/issues/2682

and for completeness: https://gitter.im/jupyter-widgets/Lobby?at=5f0ea2e30d37916fda770757

martinRenou:
Reading https://github.com/jupyter-widgets/ipywidgets/issues/2682 I feel like we have our solution! The person here seem to be able to use `IPython.display.update_display` for updating the bundle, I actually think this would be the way to go!

We would not have the issue the person mentions, simply because we would remove the widget representation so no need for re-rendering the widget, instead the UI would re-render the plot as a static image. This is what we want I guess, what do you think? :)

So I guess we can reuse the close button here, but this close button would send a figure "close" request to the Matplotlib API.
Then we need to override the closing behavior on the canvas (I hope this is possible), which would update the display bundle.

ianhi:
Would these lines https://github.com/matplotlib/ipympl/blob/64ee6af13946737ea41b2359aec8607508160eb5/ipympl/backend_nbagg.py#L294-L295 also destroy the widget object?

martinRenou:
To be totally honest with you I am not really comfortable with the Python code in this repository, it was copied from the nbagg back-end if I'm correct.

What if you override the `close` method on the Canvas and put a `print` statement in it, do you see it called when closing the figure?

ianhi:
> To be totally honest with you I am not really comfortable with the Python code in this repository,

Sadly that makes two of us


It looks like the nbagg backend (only works in notebook not in jlab) has figured out how to embed the plot as an image:
nbagg-embed

with relevant js code here: https://github.com/matplotlib/matplotlib/blob/81a560ccff1b1e369641a8687fe132d202037d4a/lib/matplotlib/backends/web_backend/js/nbagg_mpl.js#L162-L168

Button calls a js method (https://github.com/matplotlib/matplotlib/blob/81a560ccff1b1e369641a8687fe132d202037d4a/lib/matplotlib/backends/web_backend/js/nbagg_mpl.js#L53-L66) that just sets the parent node `innerHTML` to be an img tag


So that works but it seems that the discussed mime_bundle is a more principled approach

tacaswell:
If you do `display(fig)` you will get a static png (due to some other repr hooks registered by IPython).

tacaswell:
See the discussion in https://github.com/matplotlib/matplotlib/pull/17891 for more details about how Matplotlib currently interacts with the display system.

ianhi:
I spent some time messing around with rewriting the rendermime method and trying to use `update_display` but couldn't get anything working. I think this is due to widgets not playing nicely with `update_display` https://github.com/jupyter-widgets/ipywidgets/issues/1180

martinRenou:
Is there a Javascript error when trying to update the display?

jklymak:
This does not close #16 in my opinion.  Having to remember to click each figure is still a pretty big regression from `nbagg`.  Not as bad as no way to save the image, but close.

martinRenou:
Maybe I am completely missing something here, but I don't see an automatic way of updating the display for replacing the widget by a static image. So either the user will have to click the button, or they will need to programmatically close the figure.

martinRenou:
Anyway, this PR has a draft status because it simply does not work. We'll need to find a better implementation.

ianhi:
> but I don't see an automatic way of updating the display for replacing the widget by a static image. 


1. I think the ipywidgets issue with `update_display` being solved would be very helpful
2. We can maybe emulate the nbagg behavior by adding an callback to the `comm` being closed. On the comm closing `this.comm.on_close(somemethod.bind(this))` we can modify the html of the parent element to be an img tag?
    - This might require overriding `model._handle_comm_closed` https://github.com/jupyter-widgets/ipywidgets/blob/6be18d9b75353f7b4a1c328c6ea06d8959f978f6/packages/base/src/widget.ts#L202

jklymak:
> So either the user will have to click the button, or they will need to programmatically close the figure.

For nbagg, the image is just saved as part of the static representation, without user interaction.  I have no idea if that is just fundamentally impossible for widgets or not.  However, if the goal is to make a notebook available for posterity, having the static png in the notebook is a fundamental requirement.  If you don't have it, then the "widget" is just a transient interactive environment.  Which is fine, but I don't think ipympl should be touted as a replacement for `nbagg` without this ability.

martinRenou:
What makes the difference between `nbagg` and `ipympl` is the following:
- `nbagg` saves in the cell output an `application/javascript` mimetype representation that contains the code needed for when the plot is interactive, and a `text/html` mimetype representation for when the plot is static.
- `ipympl` saves in the cell output a custom ipywidgets mimetype representation which contains the widget model (and the widget manager is responsible for rendering the widget given this widget model).

We could save a `text/html` mimetype in the cell output, but that would not fix this issue, because the front-end (Jupyter Notebook or JupyterLab) will always render the widget representation if it's available, it has precedence over the static HTML representation. We actually cannot do anything about this in ipympl. That would be something to fix upstream I guess.

One thing we were trying to look at with @ianhi was to remove the widget representation and replace it with a static `text/html` representation, so either when the figure is closed programmatically or when pressing the close button. This might be doable using the IPython's `update_display`, but this does not seem to work unfortunately.

> I don't think ipympl should be touted as a replacement for nbagg without this ability.

I totally agree with this. Although to be fair, ipympl has the upside of working in JupyterLab and Voila, which is not the case for nbagg.

I would love to see this issue correctly fixed, any idea is very welcome.

martinRenou:
What we would need would be a fallback to the static HTML representation when the widget manager is not able to render the widget. That might be the perfect fix. But this is not anymore an ipympl issue then, it's an ipywidgets issue I guess.

martinRenou:
Closing as outdated. We will try another approach for saving a static representation of the plot https://github.com/matplotlib/ipympl/pull/343

PR #177, opened 2020-02-25
Move rendering logic in the Widget Model

### LINKED ISSUE #175: Front-end design issue
Currently, the widget model does not implement any logic, and the view is responsible for setting up the connection with the back-end and sending messages to it (initialization/resize etc...).

This results in some messages like the initialization one to be sent as many times as there are views, which is unwanted. 
In order to fix this, we should setup the connection in the widget model, and draw on an offscreen canvas owned by the model. Then whenever new data is received we update the offsceen canvas and update the view canvases. And the view will only be responsible for things like mouse events.

martinRenou:
This will fix #175 
This PR will make everything more robust/fast. Only the widget model is responsible for communication with the server, not each view.

It also makes the widget more stateful, by adding the header/footer/cursor as part of the widget state.

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -9,7 +9,7 @@ from IPython.display import display, HTML
 
 from ipywidgets import DOMWidget, widget_serialization
 from traitlets import (
-    Unicode, Bool, Float, List, Any, Instance, CaselessStrEnum, Enum,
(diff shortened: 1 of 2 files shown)


martinRenou:
This PR will make everything more robust/fast. Only the widget model is responsible for communication with the server, not each view.

It also makes the widget more stateful, by adding the header/footer/cursor as part of the widget state.

PR #180, opened 2020-02-26
Resize handle

### LINKED ISSUE #44: Resize handle feature
I am really missing the ability to resize the plot with the mouse like in the picture bellow with the grey bottom right corner arrow.

martinRenou:
Allow resizing the plot with the mouse, this will fix #44 

cc. @kboone 

resize

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -162,6 +162,11 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
 
     _image_mode = Unicode('full').tag(sync=True)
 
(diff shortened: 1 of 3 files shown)


martinRenou:
TODO:
- [ ] Get rid of the flickering
- [ ] Debounce the `resize` calls (do nothing if there is already a resizing happening, wait until it's finished before replaying a resize).
- [ ] Improve resize handle design?

martinRenou:
resize

jakirkham:
Thanks so much Martin for doing this work! 😄

kboone:
Thanks, this looks great! I made a couple of tweaks in #183 that I think improve usability a bit when rendering is slower.

PR #209, opened 2020-03-20
Revert postBuild change

martinRenou:
cc. @fomightez

### DIFF
--- a/postBuild
+++ b/postBuild
@@ -3,5 +3,6 @@
 # A way for mybinder.org to realize it should install the jupyter lab extension
 # automatically. This file does not affect anything except when mybinder.org use
 # it for this specifically.
-jupyter labextension install @jupyter-widgets/jupyterlab-manager jupyter-matplotlib
+jupyter labextension install @jupyter-widgets/jupyterlab-manager
+jupyter labextension install jupyter-matplotlib
 


danielballan:
Hi @martinRenou, do you understand why this is necessary?

A binder that I have had working in the past bit-rotted at some point, and this PR inspired me to try this fix https://github.com/danielballan/tutorial/commit/e0995e10039aa04cfd221f6e2995425c3558c5d4 which worked, but I do not understand why. I am near certain that consolidating to one build, as shown in linked commit, worked previously.

martinRenou:
Hey,

I cannot find the discussion we had about it @fomightez, but you told me it was not working with the one-step build? I find it odd as well.

> worked previously

Do you mean it was working making the build in one step? I usually do this locally myself and it works like a charm as well. I don't remember why @fomightez was saying it was not working on Binder.

danielballan:
Yes, if I'm not mistaken it was working doing the build in one step on Binder, but it does not work now. It still works for me locally as well.

fomightez:
Here is the discussion I think you were looking for:

https://github.com/matplotlib/ipympl/commit/236f00008892139cfedb1a47fdb61f0e87c4eef7#commitcomment-37935859

No idea though as to why.

martinRenou:
Oh thanks, I was looking for issues and PRs.

Right, so I guess we don't know :confused:

PR #213, opened 2020-04-03
Add eslint, prettier, and re-format JS

QuLogic:
This pulls in the same lint/format method used for ipywidgets, though there are a few differences to minimize changes.
* TypeScript isn't used, so I left out any typescript lint config/plugins.
* Since you've used the spread operation, I've set ECMA version to 9, though it could be set to anything later if you wish.
* Prettier is set to use 4-space tab width for existing files (except `package*.json` which is already 2-space), or else I've have to rewrite basically everything.

Some other non-defaullt changes:
* I added braces on some `if` so they would stay separate lines.
* I added a ignore pattern for args that start with `_` so `eslint` wouldn't complain they were unused.

Having a common style is helpful for cross-project sharing, and I plan to apply the same change to Matplotlib's nbagg/webagg backends. However, I am not an expert in the JS ecosystem, so I may not have worked around any issues in the best way.

Adding the config and applying the format are separate commits for easier review.

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -1,0 +1,16 @@
+name: Linting
+on: [pull_request]
+
+jobs:
+  eslint:
+    name: eslint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: eslint
+        uses: reviewdog/action-eslint@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          reporter: github-check
+          workdir: 'js'
--- a/js/.eslintrc.js
+++ b/js/.eslintrc.js
@@ -1,0 +1,21 @@
+module.exports = {
+  root: true,
+  env: {
+    browser: true,
+    commonjs: true,
+    es2017: true,
+  },
+  extends: ["eslint:recommended", "prettier"],
+  parserOptions: {
+    ecmaVersion: 9,
+    sourceType: "module",
+  },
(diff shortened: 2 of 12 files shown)


QuLogic:
I do not know what is failing here; when I run `npm install` and `npm pack` locally, it works fine.

martinRenou:
Thanks for your PR. If we can fix the CI I'd be happy to merge it.

QuLogic:
I would if I could, but it seems unrelated. There is no line with `...accumulator` in the repo.

martinRenou:
I will try to find time to look at it, sorry if it takes time

martinRenou:
Looks like this `accumulator` issue appears in other PRs so it's not related to yours. Not sure where this is coming from.

martinRenou:
@QuLogic sorry for the time it takes to merge your PR. Could you rebase your branch? The CI should pass now. It was stupid conda which was installing a very old nodejs version.

QuLogic:
Rebased. Note, this seems to have bumped the jupyter widgets version in `package-lock.json`. I'm not sure if that's okay.

QuLogic:
Actually, never mind the jupyter widgets version bump. I figured out how to get it to not change that.

martinRenou:
Awesome. Do you think it would make sense to run the lint checker in the CI? So that we make sure we respect the linter rules in the future. I would be in favor of this myself.

QuLogic:
Yes, this is something we did in Matplotlib as well; I can pull that in here, or separately.

Issue #222, opened 2020-05-09
add option to preventDefault on scroll


ianhi:
With https://github.com/matplotlib/ipympl/pull/109 the ability to respond to scroll events was brought back. However, later on the `preventDefault` was removed from the event handling code. So using https://gist.github.com/tacaswell/3144287 to implement scroll to zoom results in both zooming in and the whole notebook scrolling:
![scrolling](https://user-images.githubusercontent.com/10111092/81483220-7fd1b700-920a-11ea-9452-5785cadf92f0.gif)

Is there an easy way as a user to add back the ability to capture scroll? If not, could an option be added (default false for consistency) to capture scrolling events? 

Relevant code is here:
https://github.com/matplotlib/ipympl/blob/c4eb5127886954c038d909196e1d1b554c86bda6/js/src/mpl_widget.js#L486

I'm not sure where on the python side would be correct allow setting of an boolean for toggling capturing scroll as figure objects in other backends don't have this problem. Maybe there could be a toggle inside `fig.canvas`?

PR #235, opened 2020-06-15
Add option to preventDefault on scroll event

ianhi:
closes: https://github.com/matplotlib/ipympl/issues/222

Following the way `resizeable` was added I added `capture_scroll` to the canvas object with a default of `False`. If set to true then the entire notebook will not scroll when a user scrolls the mouse over a canvas. This allows the enduser to implement things such as custom scroll to zoom methods.

### DIFF
--- a/examples/ipympl.ipynb
+++ b/examples/ipympl.ipynb
@@ -71,6 +71,15 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "fig.canvas.capture_scroll = True"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
     "fig.canvas"
    ]
   },
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -175,6 +175,7 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
     footer_visible = Bool(True).tag(sync=True)
 
     resizable = Bool(True).tag(sync=True)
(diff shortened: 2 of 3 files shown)


dzanaga:
@martinRenou regarding #240 I tried the new feature. After installing the latest from `master`, I tried to set `fig.canvas.capture_scroll` to either `True` or `False` but it doesn't work for me. Still getting the page scrolling when trying to scroll on the canvas.

martinRenou:
Could it be an installation issue? Are you using JupyterLab or Jupyter Notebook?

dzanaga:
I am gonna try to do a fresh install of the miniconda environment when I have time and see if it's fixed there. Thanks

martinRenou:
Be sure to run `jupyter labextension install js` locally (from the ipympl root directory).

dzanaga:
That's it, it works now. Thanks for helping out and for the great work!

martinRenou:
Nice. We should make a release with this fix.

ianhi:
Also the readme currently says `jupyter labextension link ./js` and this should probably instead say `install` see https://gitter.im/jupyterlab/jupyterlab?at=5ee6abf324a3382d5d68b2ae

martinRenou:
@ianhi  If we do that, the `cd js && npm run watch` will not be relevant anymore so we might want to remove it.

ianhi:
@martinRenou instructions updated in https://github.com/matplotlib/ipympl/pull/243

> Nice. We should make a release with this fix.

:+1:  yes please

martinRenou:
@ianhi Release done on PyPi. It should come soon on conda-forge as well.

PR #245, opened 2020-07-11
Fix hidden toolbar icons

davidbailey00:
**Please read comment below and #246 before merging!**

Fixes toolbar icon(s) which were not displaying correctly due to the button width being too small to fit - just 'home' for me though this might fix #238 (presumably depends on font rendering differences between browsers / OS).

- Add the `fa-fw` class to ensure all icons are the same fixed width
- Ensure `padding: 0` is applied as this wasn't taking precedence over the default styles
  - Alternative approach would be to remove both the `width` and `padding` styles, which also leaves enough space to fit the icons, and doesn't appear (much?) different visually
- Remove unnecessary `overflow: hidden`

Before:

Home icon is missing

After:

All icons appear correctly

PR #246, opened 2020-07-12
Update dependencies

### LINKED ISSUE #238: Layout buttons not rendering correctly
I updated to the latest version of JupyterLab and `ipympl`. When plotting, some buttons don't show an icon, see figure:

widget0

## Environment
I am on linux, using miniconda3.

I used this bash script for the installation:
```
# Download latest miniconda and install
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
     bash Miniconda3-latest-Linux-x86_64.sh -b -p ${MINICONDA_PREFIX} && \
     rm Miniconda3-latest-Linux-x86_64.sh

$MINICONDA_PATH/conda init

source ${HOME}/.bashrc

# update conda to latest
conda update -y -n base conda

# configure conda to always install first from conda-forge
conda config --set channel_priority strict
conda config --add channels conda-forge

# Install default packages of the base environment
conda env update -n base -f base.yml
```

where `base.yml` is:
```
name: base
channels:
  - conda-forge
  - defaults
dependencies:
  - conda-pack
  - ipympl=0.5.6
  - matplotlib
  - nodejs
  - jupyterlab=2.1.4
  - python=3.7.6
  - pip
  - ptvsd 
  - xeus-python=0.8.0
  - pip:
    - autopep8
    - awscli==1.17.9
    - boto3==1.11.9
    - flake8
    - ipykernel
    - loguru==0.3.2
    - jupyter-lsp==0.8.0
    - jupyterlab-git
    - python-language-server[all]
    - rmate==1.0.3
```

## Versions

```bash
python -c "import sys; print('\n',sys.version); import ipympl; print('ipympl version:', ipympl.__version__)" && jupyter --version && jupyter nbextension list && jupyter labextension list
```

```
 3.7.6 | packaged by conda-forge | (default, Jun  1 2020, 18:57:50) 
[GCC 7.5.0]
ipympl version: 0.5.6
jupyter core     : 4.6.3
jupyter-notebook : 6.0.3
ipython          : 7.15.0
ipykernel        : 5.3.0
jupyter client   : 6.1.3
jupyter lab      : 2.1.4
nbconvert        : 5.6.1
ipywidgets       : 7.5.1
nbformat         : 5.0.6
traitlets        : 4.3.3
Known nbextensions:
  config dir: /home/dzanaga/miniconda3/etc/jupyter/nbconfig
    notebook section
      jupyter-matplotlib/extension  enabled 
      - Validating: OK
      nbdime/index  enabled 
      - Validating: OK
      - Validating: OK
  config dir: /usr/local/etc/jupyter/nbconfig
    notebook section
      plotlywidget/extension  enabled 
      - Validating: OK
JupyterLab v2.1.4
Known labextensions:
   app dir: /home/dzanaga/miniconda3/share/jupyter/lab
        @jupyter-widgets/jupyterlab-manager v2.0.0  enabled  OK
        @jupyterlab/debugger v0.3.0-beta.1  enabled  OK
        @jupyterlab/git v0.20.0  enabled  OK
        @krassowski/jupyterlab-lsp v1.0.0  enabled  OK
        @oriolmirosa/jupyterlab_materialdarker v0.5.0  enabled  OK
        jupyter-matplotlib v0.7.2  enabled  OK
        nbdime-jupyterlab v2.0.0  enabled  OK
```

davidbailey00:
Update dependencies to latest versions, notably:

- `eslint` v6 to v7 - none of the [breaking changes](https://eslint.org/blog/2020/05/eslint-v7.0.0-released#breaking-changes) affect the current setup of this project
- `rimraf` v2 to v3 - [drops support for Node.js 6 and older](https://github.com/isaacs/rimraf/blob/master/CHANGELOG.md#v30)
- `style-loader` v0 to v1 - none of the [breaking changes](https://github.com/webpack-contrib/style-loader/releases/tag/v1.0.0) affect the current setup of this project
- `@jupyter-widgets/base` v2 to v3 - only [dependency updates](https://github.com/jupyter-widgets/ipywidgets/compare/@jupyter-widgets/base@2.0.2...@jupyter-widgets/base@3.0.0) as far as I can tell (considering changes in the `packages/base` directory)
- `fs-extra` v4 to v9 - [various breaking changes](https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md) which are mostly bugfixes, notably Node.js v10+ is required

This seems to also fix #238 so may be worth closing #245 if this is correct - I think the newest version of `@jupyter-widgets/base` seems to inject stylesheets in the correct order consistently

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -33,20 +33,20 @@
     }
   },
   "devDependencies": {
-    "@jupyterlab/builder": "^3.0.0",
-    "css-loader": "^3.5.3",
-    "eslint": "^6.8.0",
-    "eslint-config-prettier": "^6.10.1",
-    "fs-extra": "^4.0.2",
-    "json-loader": "^0.5.4",
-    "prettier": "^2.0.2",
-    "rimraf": "^2.6.1",
-    "style-loader": "^0.23.1",
+    "@jupyterlab/builder": "^3.0.2",
+    "css-loader": "^5.0.1",
+    "eslint": "^7.19.0",
+    "eslint-config-prettier": "^7.2.0",
+    "fs-extra": "^9.1.0",
+    "json-loader": "^0.5.7",
+    "prettier": "^2.2.1",
+    "rimraf": "^3.0.2",
+    "style-loader": "^2.0.0",
     "webpack": "^5",
     "webpack-cli": "^4"
   },
   "dependencies": {
-    "@jupyter-widgets/base": "^2 || ^3 || ^4.0.0"
+    "@jupyter-widgets/base": "^4"
   },
   "keywords": [
     "jupyter",
--- a/js/package.json
+++ b/js/package.json
@@ -46,7 +46,7 @@
     "webpack-cli": "^4"
   },
   "dependencies": {
-    "@jupyter-widgets/base": "^4"
+    "@jupyter-widgets/base": "^2 || ^3 || ^4.0.0"
   },
   "keywords": [
     "jupyter",
--- a/js/src/index.js
+++ b/js/src/index.js
@@ -1,4 +1,5 @@
 // Export widget models and views, and the npm package version number.
 export * from './toolbar_widget.js';
 export * from './mpl_widget.js';
-export { version } from '../package.json';
+import pkg from '../package.json';
+export const { version } = pkg;
--- a/js/src/mpl_widget.css
+++ b/js/src/mpl_widget.css
@@ -12,7 +12,7 @@
 
 .jupyter-matplotlib-button {
   width: calc(var(--jp-widgets-inline-width-tiny) / 2 - 2px);
-  padding: 0 !important;
+  padding: 0;
 }
 
 /* Figure */
--- a/js/yarn.lock
+++ b/js/yarn.lock
@@ -44,7 +44,7 @@
     minimatch "^3.0.4"
     strip-json-comments "^3.1.1"
 
-"@jupyter-widgets/base@^4":
+"@jupyter-widgets/base@^2 || ^3 || ^4.0.0":
   version "4.0.0"
   resolved "https://registry.yarnpkg.com/@jupyter-widgets/base/-/base-4.0.0.tgz#6935461a3bd78df5523022e1c3d370ae24c9a454"
   integrity sha512-lBQgLYzq6C+XjfVJTidk+rckKo/+xlTgIm1XUtACA3BUz8bgi2du2zmbYkcrplJMwGub4QWP6GnKgM5ZZRhzYg==
--- a/js/yarn.lock
+++ b/js/yarn.lock
@@ -28,7 +28,23 @@
   resolved "https://registry.yarnpkg.com/@discoveryjs/json-ext/-/json-ext-0.5.2.tgz#8f03a22a04de437254e8ce8cc84ba39689288752"
   integrity sha512-HyYEUDeIj5rRQU2Hk5HTB2uHsbRQpF70nvMhVzi+VJR0X+xNEhjPui4/kBf3VeH/wqD28PT4sVOm8qqLjBrSZg==
 
-"@jupyter-widgets/base@^2 || ^3 || ^4.0.0":
+"@eslint/eslintrc@^0.3.0":
+  version "0.3.0"
+  resolved "https://registry.yarnpkg.com/@eslint/eslintrc/-/eslintrc-0.3.0.tgz#d736d6963d7003b6514e6324bec9c602ac340318"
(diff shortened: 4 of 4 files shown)


martinRenou:
Thanks for your PR!

Could you please revert the change on the `@jupyter-widgets/base` dependency though? We use this to allow support for JupyterLab 1 and 2. If we keep your change, only JupyterLab 2 will be supported.

js/package.json
```diff
@@ -27,18 +27,18 @@
...
+    "fs-extra": "^9.0.1",
+    "json-loader": "^0.5.7",
+    "prettier": "^2.0.5",
+    "rimraf": "^3.0.2",
+    "style-loader": "^1.2.1",
+    "webpack": "^4.43.0",
+    "webpack-cli": "^3.3.12"
   },
   "dependencies": {
-    "@jupyter-widgets/base": "^2 || ^3"
```
ianhi, line 42: I think that the `^2 || ^3` is what allows simultaneous support of jupyterlab 1 and 2.

ianhi: oh wow. i sure didn't see the other comment already saying as much before i wrote that...

davidbailey00:
Hi, sorry for leaving this for so long - I have just updated to the latest versions again and fixed merge conflicts, since this PR had since become out-of-date with the main branch :)

I just want to clarify on this:

> Could you please revert the change on the @jupyter-widgets/base dependency though? We use this to allow support for JupyterLab 1 and 2. If we keep your change, only JupyterLab 2 will be supported.

The way NPM packages work means that for `dependencies` (and `devDependencies`), the latest version which fits a range will always be used when the package is installed by a user. So for example, `^2 || ^3 || ^4.0.0` will currently resolve to `4.0.0` regardless of the version of JupyterLab the user has. Therefore, updating the range to `^4` will still work the same given that the latest version is in this range.

If this is not what you expected, it may be that `@jupyter-widgets/base` should be a `peerDependency` instead of a `dependency`. Peer dependencies are different in that the consumer of a dependency (in this case, JupyterLab) is expected to install those dependencies, rather than them being explicitly required by the package in question (i.e. this plugin).

However, if the plugin currently works as expected in JupyterLab v1 then I would advise keeping this as a `dependency` and bumping the range to only allow version 4 should make no difference :)

martinRenou:
Everything you said is true for the nbextension, the one that we build in this package and that we ship to npmjs. 

Although this is not true anymore when it comes to JupyterLab. JupyterLab has its own dependency solver (using yarn) that it uses to build itself in v1 and v2. If we enforce `@jupyter-widgets/base` v4, JupyterLab v1 and v2 will not build anymore with the ipympl labextension, because it will find a dependency conflict between `@jupyter-widgets/base`, `@jupyter-widgets/jupyterlab-manager`, `ipympl` and JupyterLab itself.

jasongrout:
In particular, JupyterLab uses yarn resolutions to force specific versions of the core packages. `@jupyter-widgets/base` depends on different versions of the core packages as well, and the major versions of `@jupyter-widgets/base` are tied to specific JupyterLab major versions. This, combined with a deduplication step that happens in a JupyterLab build, forces the JupyterLab build to choose specific versions of `@jupyter-widgets/base` for each major version of JupyterLab. That is why the version range for dependencies needs to be wide enough to allow older versions of `@jupyter-widgets/base` for older versions of JupyterLab.

jasongrout:
> If this is not what you expected, it may be that `@jupyter-widgets/base` should be a `peerDependency` instead of a `dependency`. Peer dependencies are different in that the consumer of an NPM package (the consumer being JupyterLab in this case) is expected to install those dependencies, rather than those being explicitly required by the dependency (i.e. this plugin).

Hmm, I wonder if we might be able to use peer dependencies. It seems like we should have considered this at some point, but I can't recall any specific reason we could not use peer dependencies (it's also late-I might be forgetting something). I'll think about it (or if someone else has a good answer for why peer dependencies won't work, please comment!)

martinRenou:
> It seems like we should have considered this at some point, but I can't recall any specific reason we could not use peer dependencies

This might work for the JupyterLab build to have `jupyter-widgets/base` as a peer dependency, but not for the nbextension build as it would be missing?

jasongrout:
It would need it as a dev dependency and a peer dependency, similar to how I think a lot of react packages work?

martinRenou:
Oh ok :) I didn't know about this kind of setup.

jasongrout:
I'd recommend reverting to what we know works here, and let us think through peer dependencies over in jlab and come up with a recommendation.

davidbailey00:
Thank you both for the explanations, apologies for my wrong assumption about how this works - I will revert back to using the more inclusive range

davidbailey00:
Okay, reverted to the range!

martinRenou:
Closing as outdated

Issue #108, opened 2019-07-01
Install through Extension manager


homocomputeris:
It would be nice to be able to install jupyter-matplotlib using the [Extension manager](https://jupyterlab.readthedocs.io/en/stable/user/extensions.html?highlight=extension%20manager#using-the-extension-manager).
Now jupyter-matplotlib doesn't appear there.

PR #257, opened 2020-08-14
add jupyter keywords to package.json

ianhi:
This will make jupyter-matplotlib discoverable by the jupyterlab extension manager. Which I suppose will be a totally different situation come jupyterlab3. But I imagine there will still be users on jlab2 for awhile and this will be useful to them.

Fixes: https://github.com/matplotlib/ipympl/issues/108


martinRenou (approved):

martinRenou:
Thanks!

Issue #79, opened 2019-04-27
plt.pause uses removed `start_event_loop_default` API


robtovey:
As the title suggests, attempting to pause through matplotlib currently raises an error. The minimum working example is:
```python
%matplotlib widget
from matplotlib import pyplot as plt

plt.plot([1,2,3])
plt.pause(1)    
plt.plot([3,2,1])
plt.close()
```
This raises an attribute error trying to call FigureCanvasBase.start_event_loop_default. A little googling suggests this function has been removed since [version 3.0](https://matplotlib.org/api/api_changes.html#deprecated-apis).

Taking inspiration from #17, I guess the simplest get around is using time.sleep but is there a better solution?

Thanks for your help.

ArEsKay3:
I'm not sure sleep works for me because in my case I want to pause for matplotlib event handling.

In my case I have code running in a loop that needs to periodically update the plot and get mouse events from the plot.

Is this a problem in jupyter-matplotlib or in matplotlib in general?

tacaswell:
This is an issue trying to integrate multiple event loops (see https://github.com/matplotlib/matplotlib/pull/4779 for very long discussion (I really need to finish that PR)).

`plt.pause` needs to a) block user code so that there is a 1s gap between when your two `plt.plot` calls are made, but b) not block what ever is backing your UI / background tasks.

xilunaJumma:
@robtovey Hi，I have the same problem as you ,did you have some better solution?
thanks for your help!

robtovey:
Hi @xilunaJumma, I'm afraid not. If I want to use dynamic plotting properly then I normally make a video but then I use the Agg backend anyway. Unfortunately I haven't found a particularly elegant way to replicate a 'slideshow' effect inside jupyter without having to change the backend.

jdarguello:
Same problem here. A practical solution, to run an animation using the _plt.pause()_ methodology, could be: 

1. Save the code in a python script.
2. Run the terminal in Jupyter.
3. Open the folder from where the .py script is.
4. Run the command (python 3.x): python3 your_animation_script.py

It will open a pop-up window showing the animation.

P-Mikkola:
Hello All,

[...]

Issue #258, opened 2020-08-15
Matplotlib figures are not redrawn in the cell in which they are created


kjgoodrick:
## Describe the issue

Using an interactive figure's `canvas.draw()` method does not update the figure if it was created in the same cell as the draw call.

Figure created in the same cell:
![Same cell](https://user-images.githubusercontent.com/1909532/90080052-d39ae300-dcc6-11ea-90cb-1bf3429548c6.gif)

Figure created in another cell:
![different cell](https://user-images.githubusercontent.com/1909532/90080105-02b15480-dcc7-11ea-965d-14114cb27c96.gif)

## Reproduce

1. Open a new notebook in JupyterLab or Notebook
2. Insert the following code:
````
import matplotlib.pyplot as plt
import time

%matplotlib widget

fig = plt.figure()

for i in range(5):
    x = list(range(i+2))
    xx = [x**2 for x in x]
    plt.clf()
    plt.plot(x, xx)
    fig.canvas.draw()
    
    time.sleep(1)
````
3. Run the cell and observe that the plot does not show until the loop has exited.

Optional:
1. Move the for loop to a separate cell from the figure creation
2. Run cell and observe plot updates during the loop

This can be reproduced using the [scipy notebook docker stack](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#jupyter-scipy-notebook).

## Versions

```
3.8.5 | packaged by conda-forge | (default, Jul 31 2020, 02:39:48)
[GCC 7.5.0]
ipympl version: 0.5.7
jupyter core     : 4.6.3
jupyter-notebook : 6.0.3
qtconsole        : not installed
ipython          : 7.17.0
ipykernel        : 5.3.4
jupyter client   : 6.1.6
jupyter lab      : 2.1.5
nbconvert        : 5.6.1
ipywidgets       : 7.5.1
nbformat         : 5.0.7
traitlets        : 4.3.3
Known nbextensions:
  config dir: /opt/conda/etc/jupyter/nbconfig
    notebook section
      jupyter-matplotlib/extension  enabled
      - Validating: OK
      jupyter-js-widgets/extension  enabled
      - Validating: OK
JupyterLab v2.1.5
Known labextensions:
   app dir: /opt/conda/share/jupyter/lab
        @bokeh/jupyter_bokeh v2.0.3  enabled  OK
        @jupyter-widgets/jupyterlab-manager v2.0.0  enabled  OK
[...]

PR #266, opened 2020-09-26
remove broken event loop code

ianhi:
fixes: https://github.com/matplotlib/ipympl/issues/79

A potential future ipympl specific improvement was mentioned in https://github.com/matplotlib/ipympl/issues/79#issuecomment-680077014

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -245,12 +245,6 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
     def new_timer(self, *args, **kwargs):
         return TimerTornado(*args, **kwargs)
 
-    def start_event_loop(self, timeout):
-        FigureCanvasBase.start_event_loop_default(self, timeout)
-
-    def stop_event_loop(self):
-        FigureCanvasBase.stop_event_loop_default(self)
-
 
 class FigureManager(FigureManagerWebAgg):
     ToolbarCls = Toolbar


ianhi:
@Michael-Equi can you confirm this fixes your issue. You can install with `pip install git+https://github.com/ianhi/ipympl.git@patch-1`

Michael-Equi:
That appears to have worked in eliminating the error although pause now seems to be preventing the rendering of my graph in my jupyter lab.

ianhi:
> That appears to have worked in eliminating the error although pause now seems to be preventing the rendering of my graph in my jupyter lab.

But not in notebook? Do you have a short example of what you're doing?

Michael-Equi:
I have tried running this live plot example in both Jupiter land and Jupiter notebook

```
import numpy as np 
from mpl_toolkits.mplot3d import Axes3D
from matplotlib import pyplot as plt
 
 
 
def update_line(hl, new_data):
    xdata, ydata, zdata = hl._verts3d
    hl.set_xdata(np.append(xdata, new_data[0]))
    hl.set_ydata(np.append(ydata, new_data[1]))
    hl.set_3d_properties(np.append(zdata, new_data[2]))

 
 
map = plt.figure()
map_ax = Axes3D(map)
map_ax.autoscale(enable=True, axis='both', tight=True)
 
# # # Setting the axes properties
map_ax.set_xlim3d([0.0, 10.0])
map_ax.set_ylim3d([0.0, 10.0])
map_ax.set_zlim3d([0.0, 10.0])
 
hl, = map_ax.plot3D([0], [0], [0])
 
update_line(hl, (2,2, 1))
plt.draw()
plt.show(block=False)
plt.pause(1)
 
update_line(hl, (5,5, 5))
plt.show(block=False)
plt.pause(2)
 
update_line(hl, (8,1, 4))
plt.show(block=True)
```

In both cases it does not render the plot until the final plt.show

ianhi:
maybe this is similar to: https://github.com/matplotlib/ipympl/issues/258  does the solution using async from the stackoverflow post in that issue work for you? 

As for how to make your example work I think we'll need input from someone with stronger matplotlib skills than me.

Michael-Equi:
I ended up getting it working with something like the following 

```
from matplotlib import pyplot as plt
from matplotlib.animation import FuncAnimation
from random import randrange
from threading import Thread
import time

class LiveGraph:
    def __init__(self):
        self.x_data, self.y_data = [], []
        self.figure = plt.figure()
        self.line, = plt.plot(self.x_data, self.y_data)
        self.animation = FuncAnimation(self.figure, self.update, interval=1000)
        self.th = Thread(target=self.thread_f, daemon=True)
        self.th.start()

    def update(self, frame):
        self.line.set_data(self.x_data, self.y_data)
        self.figure.gca().relim()
        self.figure.gca().autoscale_view()
        return self.line,

    def show(self):
        plt.show()

    def thread_f(self):
        x = 0
        while True:
            self.x_data.append(x)
            x += 1
            self.y_data.append(randrange(0, 100))   
            time.sleep(1)  

g = LiveGraph()
g.show()
```

martinRenou:
@ianhi is this ready to be merged?

ianhi:
I'm not sure. It removes the error in the original issue but https://github.com/matplotlib/ipympl/pull/266#issuecomment-699518908 certainly doesn't seem right and I don't think that this is expected behavior for `plt.pause`:
```python
%matplotlib widget
from matplotlib import pyplot as plt

plt.figure()
plt.plot([1,2,3])
plt.pause(2)    
plt.plot([3,2,1])
plt.close()
```
pause

compared to the same from ipython:

pause-ipython

ianhi:
So basically I think this removes the error, but doesn't fix the problem. Also, I don't know how to fix the underlying problem, or even what the underlying problem is.

martinRenou:
I see. I wonder if the `plt.pause()` does not block the comm message. It would be interesting to see if this issue happens in the nbagg backend, see if it's a regression in ipympl.

ianhi:
Notebook suffers the same affliction. 
pause-notebook

Since nbagg also doesn't implement the methods so I think we're falling back to the same thing. So it looks like merging this would make it so ipympl is even with nbagg, but both would be a regression from non-web backends.

ianhi:
I think this is related to https://github.com/matplotlib/ipympl/issues/258 which I also don't really understand.

ianhi:
Superseded by #291

martinRenou:
@ianhi really sorry for this, it didn't occur to me that it was the same PR.

ianhi:
> @ianhi really sorry for this, it didn't occur to me that it was the same PR.

No worries! I had sorta left this one hanging anyways

thomasaarholt:
Wow, I didn't even check to see if this was here! Sorry!

PR #273, opened 2020-10-14
README: Add note on npm dependency

kerel-fs:
npm is a direct dependency of the jupyter-widgets/jupyterlab-manager.

While often installed by default when installing nodejs, this must not
always be the case (e.g. [nodejs](https://www.archlinux.org/packages/community/x86_64/nodejs/) package in Archlinux doesn't contain [npm](https://www.archlinux.org/packages/community/any/npm/)).

Thus, a note on npm being a dependency is added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,8 @@ pip install ipympl
 
 ### Install the JupyterLab extension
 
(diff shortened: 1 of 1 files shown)


martinRenou:
Thanks!

Issue #275, opened 2020-10-20
%matplotlib widget not rendering


blackbox-tech:
I'm trying to switch from ```jupyter notebook``` to ```jupyter lab``` but I can't get the interactive plots in my notebooks working.

Instead of rendering the plot (in Firefox or Chrome) it just prints something like this:
```
Canvas(toolbar=Toolbar(toolitems=[('Home', 'Reset original view', 'home', 'home'), ('Back', 'Back to previous \u2026
```
From reading issues posted by other users, it looks like a missing package dependency in conda, but I can't figure out which package is missing.  I spend a few hours trying to get it working from clean conda envs, but I have had no luck, please advise.

After installing various packages to try to workaround this issue, this is what is in my currently in conda environment:
```
conda list
# packages in environment at /home/paul/.conda/envs/test_new:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       1_gnu    conda-forge
argon2-cffi               20.1.0           py39h07f9747_2    conda-forge
async_generator           1.10                       py_0    conda-forge
attrs                     20.2.0             pyh9f0ad1d_0    conda-forge
backcall                  0.2.0              pyh9f0ad1d_0    conda-forge
backports                 1.0                        py_2    conda-forge
backports.functools_lru_cache 1.6.1                      py_0    conda-forge
bleach                    3.2.1              pyh9f0ad1d_0    conda-forge
brotlipy                  0.7.0           py39h38d8fee_1001    conda-forge
ca-certificates           2020.6.20            hecda079_0    conda-forge
certifi                   2020.6.20        py39h079e4ff_2    conda-forge
cffi                      1.14.3           py39h0da4893_1    conda-forge
chardet                   3.0.4           py39h079e4ff_1008    conda-forge
configurable-http-proxy   1.3.0                         0    conda-forge
cryptography              3.1.1            py39h3c970e7_1    conda-forge
cycler                    0.10.0                     py_2    conda-forge
decorator                 4.4.2                      py_0    conda-forge
defusedxml                0.6.0                      py_0    conda-forge
entrypoints               0.3             py39hde42818_1002    conda-forge
freetype                  2.10.4               he06d7ca_0    conda-forge
icu                       67.1                 he1b5a44_0    conda-forge
idna                      2.10               pyh9f0ad1d_0    conda-forge
importlib-metadata        2.0.0                      py_1    conda-forge
[...]

Issue #201, opened 2020-03-13
a simple interactive matplotlib plot is not showed in jupyter lab


eurydice76:
Hello,

I try to plot interactively a simple plot through jupyterlab. Here is the code I use to do so:

```
%matplotlib widget
plt.figure()
x = [1,2,3]
y = [4,5,6]
plt.plot(x,y)
plt.show()
```

Unfortunately, running that code produces the following output:

`Canvas(toolbar=Toolbar(toolitems=[('Home', 'Reset original view', 'home', 'home'), ('Back', 'Back to previous …
`

which seems to indicate that something is wrong with the rendering.
I followed the instructions for installing all the dependencies according other posts and documentation but it did not help. Here is what I installed:

```
conda install -c conda-forge ipympl
conda install -c conda-forge nodejs
conda install -c conda-forge widgetsnbextension
jupyter labextension install @jupyter-widgets/jupyterlab-manager
jupyter labextension install jupyter-matplotlib
```
Here is some info about the versions of my different packages:

```
python: 3.6.7
ipympl: 0.5.3
jupyter-lab: 2.0.1
jupyterlab-extension: 2.0.1
jupyter core     : 4.6.3
jupyter-notebook : 6.0.3
qtconsole        : not installed
ipython          : 7.13.0
ipykernel        : 5.1.4
jupyter client   : 6.0.0
jupyter lab      : 2.0.1
nbconvert        : 5.6.1
ipywidgets       : 7.5.1
nbformat         : 5.0.4
traitlets        : 4.3.3
```
would you have any idea ?

Many thanks in advance

martinRenou:
You did not miss anything in the installation. 

Did you refresh the JupyterLab page?

eurydice76:
yeap, I refreshed the page, logout from jupyterhub but still the same bad behaviour

martinRenou:
Do you get an error in the JavaScript console? `ctrl + shift + J` on Chrome, `ctrl + shift + K` on Firefox

eurydice76:
the only "Error" I can see is the following:

`Error: "Object 'jupyter.widget' not found in registry"`

martinRenou:
Ok. Did you try rerunning `jupyter labextension install @jupyter-widgets/jupyterlab-manager jupyter-matplotlib`? There was a period of time when this did not work around last week.

eurydice76:
I launched those commands just a few hours ago

martinRenou:
[...]

Issue #148, opened 2019-10-27
%matplotlib widget not working in jupyter lab


hasibzunair:
Running a cell in jupyter lab with `%matplotlib widget` results in: 

```Canvas(toolbar=Toolbar(toolitems=[('Home', 'Reset original view', 'home', 'home'), ('Back', 'Back to previous …```
Running the cell more than once results in: 

```KeyError: '97acd0c8fb504a2288834b349003b4ae'```

My environment:
* Python 3.6.8
* Matplotlib 3.0.3

**Comment by martinRenou** [comment]
What instructions did you follow for installing `ipympl`?

**Comment by hasibzunair** [comment]
Just `pip install ipympl`.

It works now after running the following bash commands.
```
conda install -y nodejs
pip install --upgrade jupyterlab
jupyter labextension install @jupyter-widgets/jupyterlab-manager
jupyter labextension install jupyter-matplotlib
jupyter nbextension enable --py widgetsnbextension
```

**Comment by Nestak2** [comment]
Thanks, "restart" was the critical instruction for me. It wasn't enough the quit and start again jupyterlab in the terminal, only after restarting the computer the plots finally worked

**Comment by martinRenou** [comment]
Usually clearing the browser cache is the important step

**Comment by RGD2** [comment]
FWIW I had same problem, but had to avoid using conda, and was required to use virtual envs with python. 
I was able to fix this by following [nodejs's install instructions using a package manager](https://nodejs.org/en/download/package-manager/). In my case this was ubuntu -- so [using nodejs binary distributions from github](https://github.com/nodesource/distributions/blob/master/README.md).

This ended up being something like (in total):
```
mkvirtualenv -p python3.8 jupenv
pip install jupyter matplotlib pandas ipympl tqdm
curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash -
sudo apt-get install -y nodejs
jupyter labextension install @jupyter-widgets/jupyterlab-manager
```

Because of the venv, you later start jupyter with:

    workon jupenv
    jupyter lab

Hope this helps others.

**Comment by Mariveros** [comment]
Same problem here, restarting kernel not work for me. only when I restart jupyter lab from terminal it works. (%matplotlib widget)

**Comment by Samadarshi-Maity** [comment]
upgrade pip
add 
```
pip install ipympl 
```
in anaconda prompt
and restart jupyter notebook

Issue #66, opened 2018-09-12
Interactivity does not work in Jupyter Lab


lschr:
Everything works fine in Jupyter Notebook, but not in Jupyter Lab (version 0.34.7).

![screenshot](https://user-images.githubusercontent.com/13418822/45416986-6a953580-b681-11e8-8d4e-d36942fc0836.png)

- Styling of the buttons does not seem to work
- The buttons have no functionality; e.g. pressing the panning button does not enable panning

On the Javascript console I can the following error whenever the mouse pointer enters the area of the figure, which may or may not be related:

![screenshot_20180912_130449](https://user-images.githubusercontent.com/13418822/45421134-6f131b80-b68c-11e8-89e0-5048c568f859.png)

feanor12:
I had a similar problem when I was using a notebook which had output cells with plots generated by (%matplotlib notebook)

I fixed it by:
1. delete all output cells
2. save notebook
3. reload page
4. restart kernel 
5. run with %matplotlib widget

I think that %matplotlib notebook injects some kind of javascript that kills something inside jupyter lab.

lschr:
@feanor12 Thanks, it looks like you are right. After restarting and clearing all output, the buttons work (except for the "power" button, which has already been reported in #61) and also the mouse pointer coordinates are shown properly in the bottom right corner.

The buttons still look Windows 95-ish, though (not sure, if this is a bug).

feanor12:
So maybe this is the cause for #62 as well.

MerlinPendragon:
@feanor12 Yes!

pdeitel:
Is this supposed to work with Matplotlib FuncAnimations too? All I get is a blank output and the FuncAnimation itself--which works just fine in IPython in a Terminal window--does not call the update function I defined for each frame.  Thanks in advance for any suggestions.

tacaswell:
@pdeitel that is extremely odd.  I am a bit worried that your generator is raising someplace and the exceptions are getting snarfed incorrectly.  Internally, the animation module homgonizes all the input to generators (an integer gets turned into `range`).

tacaswell:
> I think that %matplotlib notebook injects some kind of javascript that kills something inside jupyter lab.

Yes, all of the js to make the plot interactive is inserted by the notebook backend (which is why it does not work in jlab as they have closed that convenient but large security hole).

goiosunsw:
[...]

PR #279, opened 2020-11-23
Add instruction to restart jupyterlab

orena1:
All of those are because they did not restart JupyterLab
https://github.com/matplotlib/ipympl/issues/275
https://github.com/matplotlib/ipympl/issues/148
https://github.com/matplotlib/ipympl/issues/66
https://github.com/matplotlib/ipympl/issues/201

Happen to me also, might even make people go back to notebook

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ jupyter labextension install @jupyter-widgets/jupyterlab-manager
 # If you already installed the @jupyter-widgets/jupyterlab-manager extension, you will still need to rebuild JupyterLab after you installed ipympl
 jupyter lab build
 ```
+Make sure to restart JupyterLab after installing the extension
 
 #### Install an old JupyterLab extension
 


jasongrout:
After rebuilding, you should just need to refresh your browser. Restarting forces a refresh, so that works too.

orena1:
I tried  refreshing a couple of time, it did not work. I still got:
`Canvas(toolbar=Toolbar(toolitems=[('Home', 'Reset original view', 'home', 'home'), ('Back', 'Back to previous`

jasongrout:
> I tried refreshing a couple of times, it did not work. I still got:

But restarting did work? That output indicates `@jupyter-widgets/jupyterlab-manager` is not installed/built.

orena1:
Yes, restart did work. 
It is not that it completely did not work completely before restart. If I would do display(fig) it worked, but just not automatically for each figure that I created. 
For example something like this:
```python
# Testing matplotlib interactions with a simple plot
figure()
imshow(mov0[50])
figure()
imshow(mov1[50])
figure()
imshow(mov2[50])
```

did not work before I restarted.

jasongrout:
Ah, I'm also presuming that you restarted your kernel after installing the python package. A restart would also force that.

orena1:
I tried both kernel restart and browser refresh and it did not work.
Only when I restarted JupyterLab it started to work correctly. 
This is also the advise that solved it for these ticket:
https://github.com/matplotlib/ipympl/issues/275#issuecomment-713016580
https://github.com/matplotlib/ipympl/issues/148#issuecomment-546994990

jasongrout:
As far as I'm aware, a complete server restart shouldn't be necessary in general, thought it certainly is sufficient. Maybe there is something more tricky with how things work with ipympl. Anyways, I'm glad you are contributing back what worked for you, and it's up to the maintainers here to review/decide.

martinRenou:
👍🏽

PR #283, opened 2021-01-04
add regular test against maptlotlib master

ianhi:
Runs nbval with the latest matplotlib master branch in order to catch breaking changes prior to the matplotlib releases. 

Runs both on PRs and every monday at noon UTC. 

cc @martinRenou


martinRenou:
Thanks!

PR #284, opened 2021-01-04
Add automated release workflow

ianhi:
This would allow the building of the wheel and the uploads to pypi and NPM to happen automatically on making a new release on github.

If this gets merged then someone with rights to publish to those platforms will need to generate an API key and add it to this repos secrets.

again @martinRenou  (sorry for sudden spam :/) but i'm hoping this one can make your life a bit easier when releasing :)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,38 @@
+name: Publish Package
+
+on:
+  release:
+    types: [published]
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install node
+      uses: actions/setup-node@v1
+      with:
+        node-version: '14.x'
+        registry-url: 'https://registry.npmjs.org'
+    - name: Install Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install packaging setuptools twine wheel
+    - name: Publish the Python package
+      env:
+        TWINE_USERNAME: __token__
+        TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload dist/*
+    - name: Publish the NPM package
+      run: |
+        echo $PRE_RELEASE
+        if [[ $PRE_RELEASE == "true" ]]; then export TAG="next"; else export TAG="latest"; fi
+        npm publish --tag ${TAG} --access public
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        PRE_RELEASE: ${{ github.event.release.prerelease }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,16 +20,17 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install packaging setuptools twine wheel
+        pip install twine wheel build
     - name: Publish the Python package
       env:
         TWINE_USERNAME: __token__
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install packaging setuptools twine wheel
+    - name: Publish the Python package
+      env:
+        TWINE_USERNAME: __token__
+        TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
```
martinRenou, line 27: How are we supposed to generate and set this token?

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload dist/*
+    - name: Publish the NPM package
+      run: |
+        echo $PRE_RELEASE
+        if [[ $PRE_RELEASE == "true" ]]; then export TAG="next"; else export TAG="latest"; fi
+        npm publish --tag ${TAG} --access public
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
martinRenou, line 37: Same question here I guess :)

martinRenou:
Thanks for the PR!

.github/workflows/release.yml
ianhi: For generating go to pypi-> account settings then scroll down to:
image

For setting you add it to the `Secrets` section of this repo in settings like so:
image


then it all sorta "just works"

.github/workflows/release.yml
ianhi: same as pypi except token created on npm.js instead of pypi

ianhi:
Once the tokens are set then the release process becomes:
1. Draft a new release via github UI
2. Publish that release

and it will even do the git tagging for you

ianhi:
This is a near carbon copy of the juptyerlab-git version https://github.com/jupyterlab/jupyterlab-git/blob/master/.github/workflows/publish.yml

martinRenou:
I like more and more the idea of doing this. I would like to test it on a less-critical project first though. I might start with ipycanvas.

ianhi:
> I would like to test it on a less-critical project first though. I might start with ipycanvas.

Sensible. Feel free to just copy paste this PR over to there. Alternatively we can try it on ipycytoscape. I would have done that already (https://github.com/QuantStack/ipycytoscape/blob/master/.github/workflows/release.yml) but I don't have adequate permissions to add secrets to the repo

martinRenou:
Me neither. Maybe @marimeireles has those rights?

martinRenou:
I am not allowed to set up secrets on ipympl. Which makes sense.

@tacaswell would you be able to create the secret tokens for publishing to PyPi and npmjs?

I'll do the next release manually.

marimeireles:
Sorry, I missed this.
I just checked I don't have permission to add secrets.
@SylvainCorlay adding these secrets would automate the release of ipycytoscape a bit, could you give these permissions to Ian or me?

ianhi:
Hi @tacaswell would you be ok with adding the necessary secrets? This would make it super easy to do things like release immediately after merging: https://github.com/matplotlib/ipympl/pull/310

Also fwiw `pypi` token can be scoped by package, but `npmjs` cannot and give access to all packages.

ianhi:
Automated releases have been tested out for both ipycanvas and jupyterlab-sidecar now and seems to work for both.

tacaswell:
I suspect I have access to add the secrets on the GH side, but do not have access to get them on the pypi side!

That said, if I can, I am happy to pull levers and push buttons as required.

ianhi:
The tests won't detect anything here. So going with a merge. Only way to check is to try releasing.

Issue #290, opened 2021-01-06
What prevents a figure showing before notebook cell completion?


thomasaarholt:
My issue boils down to:

"Why does the figure not show immediately, but instead waits until the cell has completed running?"
```python
%matplotlib widget
import matplotlib.pyplot as plt
from time import sleep
fig = plt.figure()
plt.plot([1,2,3])
sleep(2)
```
<img src="https://user-images.githubusercontent.com/2721423/103765666-d43d6880-501d-11eb-8ef7-b1b15bb5cfae.gif" width="300px" alt="Figure only showing at end of sleep/cell">

This behaviour is preventing fun uses, like dynamic plotting in a single cell:
```python
%matplotlib widget
from matplotlib import pyplot as plt
from time import sleep
import numpy as np
fig = plt.figure()

for i in range(10):
    x, y = np.random.random(2)
    plt.scatter(x, y)
    fig.canvas.draw()
    sleep(0.1)
```

The previous code works if one splits it into two cells, placing the `figure()` call in the first one:

```python
# Cell 1
%matplotlib widget
from matplotlib import pyplot as plt
from time import sleep
import numpy as np
fig = plt.figure()

# Cell 2
for i in range(10):
    x, y = np.random.random(2)
    plt.scatter(x, y)
    fig.canvas.draw()
    sleep(0.1)
```

<img src="https://user-images.githubusercontent.com/2721423/103765422-74df5880-501d-11eb-84c9-19efc5172ec0.gif" width="300px" alt="Dynamic plotting if using two cells">

thomasaarholt:
For anyone who is looking for a temporary solution, I'll add that for my project, where I am generating figures to be able to keep track of whether an image correction algorithm is proceeding in the right direction, I'm currently creating a figure in the notebook, and then getting hold of that in my code library with `plt.gcf()`, so the general workflow is:

```python
# Cell 1
plt.figure(figsize=some_size_for_a_nice_aspect_ratio)

# Cell 2
while not_converged:
    # (...)
    fig = plt.gcf()
    fig.clear()
    ax = fig.add_subplot()
    ax.plot(...)
```

ianhi:
My understanding of this is that:
1. Running python code blocks the processing of comm messages
2. At least one incoming comm message needs to be processed by the python side in order for the js side to finish initialization
3. ergo python execution must finish to allow the comms and the backend to reply to the frontend.

In particular the issue is due to the dpi handling that's done in the initialization method on js side:
[...]

PR #291, opened 2021-01-06
Fix plt.pause() behaviour

thomasaarholt:
In #79, it is shown that plt.pause() raises an error when using ipympl.
```python
%matplotlib widget
fig = plt.figure()
plt.plot([1,2,3]) # this line isn't necessary
plt.pause(1)    
# AttributeError
```

In that issue, there is discussion recommending removing the two methods that raise the error, as the `FigureCanvasBase.start_event_loop_default` method does not exist. This PR implements that.

It turns out that the ipympl Canvas object actually inherits from `FigureCanvasBase` anyway (through `Canvas` -> `FigureCanvasWebAggCore` -> `backend_agg.FigureCanvasAgg` -> `FigureCanvasBase`), so removing them or renaming the called methods to remove the `_default` are equivalent.

Here is a gif showing the new behaviour with this PR (previous was an error)

Working plt.pause()

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -242,12 +242,6 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
     def new_timer(self, *args, **kwargs):
         return TimerTornado(*args, **kwargs)
 
-    def start_event_loop(self, timeout):
-        FigureCanvasBase.start_event_loop_default(self, timeout)
-
-    def stop_event_loop(self):
-        FigureCanvasBase.stop_event_loop_default(self)
-
 
 class FigureManager(FigureManagerWebAgg):
     ToolbarCls = Toolbar
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -19,8 +19,7 @@ from matplotlib.backends.backend_webagg_core import (FigureManagerWebAgg,
                                                      FigureCanvasWebAggCore,
                                                      NavigationToolbar2WebAgg,
                                                      TimerTornado)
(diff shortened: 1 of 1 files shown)


martinRenou:
Thanks a lot! Do you think it's also related to the other issue you opened (https://github.com/matplotlib/ipympl/issues/290)? 

There is a small flake8 issue `F401 'matplotlib.backend_bases.FigureCanvasBase' imported but unused` otherwise happy to merge it and make a patch release.

thomasaarholt:
I removed that import now 👍  (Edit: I see that you merged - thanks!)

This PR is related to #291, but only in fixing the behaviour of plt.pause() so that it works like `sleep()` does.

Do you have any idea where the "blocking" behaviour in #291 sits? I was trying to think how qt does it. :/

martinRenou:
> Do you have any idea where the "blocking" behaviour in #291 sits? I was trying to think how qt does it. :/

Do you mean #290? Not sure, it looks related to other issues that are open. I think @ianhi looked at something similar concerning blocking comms or something. I suspect the comms are stuck and queued due to Matplolib, but I would need to look deeper into the code, I have to admit I never investigated this issue.

thomasaarholt:
Sorry, yes #290. It's something that I think would be really nice - currently I'm tracking whether an algorithm is converging by generating plots as it runs, but I'd have to pre-generate all the figures for it to work with ipympl.

mforbes:
@thomasaarholt In your example at the top, Cell #2 does not produce output., whereas when I run it now (matplotlib=3.10.1, ipympl=0.9.7) , the second cell produces 10 plots unless I use `sleep(0.1)` instead of `plt.pause(0.1)` as you have.  I have added output in #290.

PR #294, opened 2021-01-16
Embed image on save

martinRenou:
This is an attempt to embed the Image state on Notebook save (JupyterLab only for now, we can maybe do something similar in Jupyter Notebook classic).

The logic here is to:
- Hook up to the "save" event of the Notebook
- Go through all outputs and find matplotlib widgets
- Replace the widget output into plain html image

See https://github.com/matplotlib/ipympl/issues/150 https://github.com/matplotlib/ipympl/issues/16


martinRenou:
This doesn't work yet. I am opening the PR as a draft for heads up/

martinRenou:
Might be replaced by #300

martinRenou:
Closing as outdated. We will try another approach for saving a static representation of the plot #343

PR #300, opened 2021-02-11
Implement generateMimebundle

martinRenou:
Builds on top of https://github.com/jupyter-widgets/ipywidgets/pull/3107

### DIFF
--- a/js/src/mpl_widget.js
+++ b/js/src/mpl_widget.js
@@ -241,10 +241,31 @@ export class MPLCanvasModel extends widgets.DOMWidgetModel {
     }
 
     generateMimeBundle() {
-        return Promise.resolve({
-            'text/html': `<img src="${this.offscreen_canvas.toDataURL('image/jpeg', 1.0)}"
-                           width="${this.offscreen_canvas.width}" height="${this.offscreen_canvas.height}"/>`
-        });
+        const width = this.offscreen_canvas.width;
+        const height = this.offscreen_canvas.height;
+
+        let header = '';
+        if (this.get('header_visible')) {
+            header = `
+                <div style="text-align: center;">${this.get('_figure_label')}</div>
+            `
+        }
+
+        return {
+            'text/html': `
+                <div>
+                    ${header}
+                    <img src="${this.offscreen_canvas.toDataURL('image/png', 1.0)}"
+                    width="${width}" height="${height}"
+                    style="width: ${width / this.ratio}px; height: ${height / this.ratio}px;"/>
+                </div?
+            `
+        };
+    }
+
+    // We want the static figure to overwrite the widget mimebundle on Notebook save
+    shouldOverwriteMimeBundle() {
+        return true;
     }
 
     remove() {
--- a/js/src/mpl_widget.js
+++ b/js/src/mpl_widget.js
@@ -240,6 +240,13 @@ export class MPLCanvasModel extends widgets.DOMWidgetModel {
         }
     }
 
+    generateMimeBundle() {
+        return Promise.resolve({
+            'text/html': `<img src="${this.offscreen_canvas.toDataURL('image/jpeg', 1.0)}"
+                           width="${this.offscreen_canvas.width}" height="${this.offscreen_canvas.height}"/>`
+        });
+    }
+
     remove() {
         this.send_message('closing');
     }
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ js/*.tgz
 
 # OS X
 .DS_Store
+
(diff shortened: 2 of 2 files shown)


ianhi:
Something I meant to comment on #294: Will this stop widget interaction on a save event? Ideally the interaction will stay open on saving. I know that I regularly save as a matter of habit so that would be a pretty big disruption.

Is it the case that this mimebundle will only be displayed if the model state is not available?

ianhi:
also: hoooooooooooooooorayyyyyyyyyyyyy - I'm so happy y'all are making progress on this

js/src/mpl_widget.js
```diff
@@ -240,6 +240,13 @@ export class MPLCanvasModel extends widgets.DOMWidgetModel {
         }
     }
 
+    generateMimeBundle() {
+        return Promise.resolve({
+            'text/html': `<img src="${this.offscreen_canvas.toDataURL('image/jpeg', 1.0)}"
```
ianhi, line 245: why `jpeg`? FWIW the backend provides a PNG. The data comes from here:
https://github.com/matplotlib/matplotlib/blob/56d9e7cb1718b740dee717ac0a208f7eae9db204/lib/matplotlib/backends/backend_webagg_core.py#L239-L242

martinRenou: IDK actually, it's the first idea that came to my mind. We can do PNG then :)

martinRenou:
> Something I meant to comment on #294: Will this stop widget interaction on a save event? Ideally the interaction will stay open on saving. I know that I regularly save as a matter of habit so that would be a pretty big disruption.

It seems like saving the Notebook does not change the content of the Notebook when looking at it. The widget stays there and doesn't get re-rendered.

martinRenou:
@ianhi the fact that [Maarten 's PR](https://github.com/jupyter-widgets/ipywidgets/pull/3107) does not overwrite the widget state from the mimebundle will be a blocker in the case of ipympl.

Given the current implementation of ipympl, even if we have a widget model, we are not able to recover the plot image (because it's not part of the model).

Also, even if we made the plot image part of the model, when opening a Notebook saved by someone else, the widget might not be interactive anyway.

I would be in favor of adding a custom save hook in ipympl for Jupyter Notebook, that would remove the widget repr from the bundle, and only save the static image. What do you think?

sin-mike:
> @ianhi the fact that [Maarten 's PR](https://github.com/jupyter-widgets/ipywidgets/pull/3107) does not overwrite the widget state from the mimebundle will be a blocker in the case of ipympl.
> 
> Given the current implementation of ipympl, even if we have a widget model, we are not able to recover the plot image (because it's not part of the model).
> 
> Also, even if we made the plot image part of the model, when opening a Notebook saved by someone else, the widget will not be interactive anyway, because the new back-end is not aware of the data it should plot if we interact with the buttons.
> 
> I would be in favor of adding a custom save hook in ipympl for Jupyter Notebook, that would remove the widget repr from the bundle, and only save the static image. What do you think?

Dear @martinRenou , is there any resolution to the issue yet? I'm really looking forward to the working raster previews saved in NBs. If not, what would you suggest as a temporary workaround for now?

Sincerely, 
Mike

martinRenou:
cc. @ianhi 

Changing https://github.com/jupyter-widgets/ipywidgets/pull/3107 so that it adds the possibility to completely overwrite the mimebundle gives the most wanted result:

ipympl_sav

Will push the commits soon-ish.

martinRenou:
New update

ipympl_save

martinRenou:
> If not, what would you suggest as a temporary workaround for now?

@sin-mike there is no workaround possible as far as I know. But I am hopeful we can fix this soonish.

ianhi:
> Dear @martinRenou , is there any resolution to the issue yet? I'm really looking forward to the working raster previews saved in NBs. If not, what would you suggest as a temporary workaround for now?

@sin-mike it's very much not ideal but if you get a figure to where you want it then there are two options to embed a static version in the notebook:
1. `display(fig)`
2. Screenshot/save and paste into markdown cell (if screenshot goes to clipboard you can directly paste into markdown cell)

---

@martinRenou  looking great. I imagine most of the work for this will live on the ipywidgets pr?

martinRenou:
> @martinRenou looking great. I imagine most of the work for this will live on the ipywidgets pr?

Yeah. After some discussions with @SylvainCorlay we agreed that this was too much of a hack. We're trying a cleaner approach.

sin-mike:
>     1. display(fig)
Wow, it's really a quick way to rasterize plots. Thanks!

Issue #303, opened 2021-02-23
ipympl version 0.6.3 is incompatible with matplotlib 3.3.4


kthare10:
## Describe the issue
ipympl version 0.6.3 is incompatible with matplotlib 3.3.4

## Exception
draw() throws an exception shown below when trying to access `_activeQue` in matplotlib/_pylab_helpers.py. This member only exists in matplotlib version 3.2.2 and has been removed in version 3.3.*
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-a2dfa485d491> in <module>
      7 t.nodes['n2'].add_component(ctype=fu.ComponentType.SharedNIC, model='ConnectX-6', name='nic2')
      8 t.add_link(name='l1', interfaces=list(t.interfaces.values()), ltype=fu.LinkType.Wave)
----> 9 t.draw()
     10 t.draw(topo_detail=fu.TopologyDetail.Derived)
/opt/conda/lib/python3.8/site-packages/fim/user/topology.py in draw(self, file_name, interactive, topo_detail, layout)
    300             nx.draw_networkx(derived_graph, pos=pos)
    301             if not interactive:
--> 302                 plt.show()
    303             if file_name is not None:
    304                 plt.savefig(file_name)
/opt/conda/lib/python3.8/site-packages/matplotlib/pyplot.py in show(*args, **kwargs)
    351     """
    352     _warn_if_gui_out_of_main_thread()
--> 353     return _backend_mod.show(*args, **kwargs)
    354 
    355 
/opt/conda/lib/python3.8/site-packages/ipympl/backend_nbagg.py in __call__(self, block)
     54                 manager.canvas.mpl_disconnect(manager._cidgcf)
     55 
---> 56             if not interactive and manager in Gcf._activeQue:
     57                 Gcf._activeQue.remove(manager)
     58 
AttributeError: type object 'Gcf' has no attribute '_activeQue'
```

## Versions
python -c "import sys; print('\n',sys.version); import ipympl; print('ipympl version:', ipympl.__version__)" && jupyter --version && jupyter nbextension list && jupyter labextension list
 3.8.6 | packaged by conda-forge | (default, Dec 26 2020, 05:05:16) 
[GCC 9.3.0]
ipympl version: 0.6.3
jupyter core     : 4.7.1
jupyter-notebook : 6.2.0
qtconsole        : not installed
ipython          : 7.20.0
ipykernel        : 5.4.2
[...]

PR #305, opened 2021-03-01
Use _Backend from matplotlib to simplify the code

ianhi:
Follows how the nbagg backend currently handles things: https://github.com/matplotlib/matplotlib/blob/04f78b84a3fda987b4ac65da296d7e4aa764d894/lib/matplotlib/backends/backend_nbagg.py#L225

Conversation with @tacaswell here https://gitter.im/matplotlib/matplotlib?at=603c83fc44f5a454a447c6e9 left me with the impression that it is ok to use the private class.

This ended up being most just updating the code to match nbagg, but my taking the new show function this also fixes: https://github.com/matplotlib/ipympl/issues/303

I also created UAT.ipynb for user automated tests. This is helpful for things like the issue in #303 that are important to test as they may break again but don't really make sense as an example.


ianhi:
Perhaps using `_Backend` is not optimal see: https://github.com/matplotlib/matplotlib/issues/19605#issuecomment-787919282

Though I still think it is currently the best solution

martinRenou:
Do you think we should put `UAT.ipynb` under a test directory?

ianhi:
> Do you think we should put UAT.ipynb under a test directory?

I ceated a top level `test-notebooks` directory because it doesn't feel like it would belong in a proper pytest `tests` directory, but happy to change.

I also moved the pytest settings into pyproject.toml to consolidate files.

ianhi:
Bump of this @martinRenou and also @tacaswell in case this is something that ipympl really should **not** be doing?

I think having a UAT notebook is very a nice thing, but we can also swap that into a new PR if need be.

.gitignore
```diff
@@ -13,3 +13,6 @@ js/*.tgz
 
 # OS X
 .DS_Store
+
+# vscode
+.vscode/
```
tacaswell, line 18:
```suggestion
.vscode/

```

tacaswell (approved):
modulo my cranky end with a new line suggestion

ianhi:
I think the LGTM has stalled out, and it was green before adding the newline so going for it

PR #310, opened 2021-03-17
FIX: port the key handling changes from mpl

### LINKED ISSUE #309: incompatibility with mpl 3.4
## Describe the issue

https://github.com/matplotlib/matplotlib/pull/19146 made changes to how we encode keyboard events on the js side to ship back to the Python side.  

This PR greatly improved the uniformity of handling the keyboard, but think we need to port the js side changes of https://github.com/matplotlib/matplotlib/pull/19146/files?file-filters%5B%5D=.js#diff-179115fdfd653499287a838c855c18302928eccf4d4878acb93da98701246d0d to ipympl or locally restore the old implementation of `_handle_key`.

tacaswell:
ports changes from matplotlib/matplotlib#19146

closes #309

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -12,6 +12,7 @@ from traitlets import (
     default
 )
 
+import matplotlib
 from matplotlib import rcParams
 from matplotlib.figure import Figure
 from matplotlib import is_interactive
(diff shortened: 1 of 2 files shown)


tacaswell:
Style checkers: 2 Tom: nil

ipympl/backend_nbagg.py
```diff
@@ -241,6 +242,68 @@ def send_binary(self, data):
     def new_timer(self, *args, **kwargs):
         return TimerTornado(*args, **kwargs)
 
+    if matplotlib.__version__ < '3.4':
```
ianhi, line 245: woah neat! I had no idea this was a thing you can do

ianhi:
:+1: Makes sense to me. 

It would also be great to have somewhere to put your test case https://github.com/matplotlib/ipympl/issues/309#issuecomment-801483767 somewhere (especially for whenever a proper widget testing framework solidifies - which I think will be based on https://github.com/jupyterlab/galata ???) 

A good place would be the UAT notebook from https://github.com/matplotlib/ipympl/pull/305. But if this gets merged first lets just make sure to open a follow up issue for adding to the UAT.

martinRenou:
> which I think will be based on https://github.com/jupyterlab/galata ???

That would be great indeed.

tacaswell:
I'm not sure what the current ipympl release schedule is, but can this go out as soon as possible?  We are probably going to do mpl 3.4 in the next few days.

ianhi:
> I'm not sure what the current ipympl release schedule is, but can this go out as soon as possible? We are probably going to do mpl 3.4 in the next few days.

I think the release schedule has been: Something gets merged, then we wait a few days until someone remembers to make a new release. So lets release right after merging this. (Or maybe after this + https://github.com/matplotlib/ipympl/pull/305)

ianhi:
I tried this out, but it's not actually clear to me what this fixes? Is it only possible to see the bug in a non-qwerty layout?

Overall I trust that this fixes the issue though

ianhi (approved):

tacaswell:
Without this patch, but with mpl3.4.0rcX / v3.4.x / master when you type keys you get back the key code as a string.  mpl3.3 and before we shipped back the key codes and then converted to keys a (private) method on the Python side.  We overhauled that code for mpl34 which required changes to both the py and js sides. ipympl re-uses the Python side logic (did not trace it through fully, but it is via inheritance) but has its own version of the js side.

This PR back-ports the js side, so that we feed mpl34 the values it expects, but that would break ipypmpl with any, ah, currently released version of Matplotlib which seems bad.  So this also backports the conversion function and hides it behind a mpl version-gated at class-definition time check to selectively over-ride the private method.  Thus if we do change the private function again (without changing the protocol to with the js side), then ipympl will get the improvements and old versions of mpl still work :)

martinRenou:
Thanks!

ianhi:
This should be in version `0.7.0` on pypi. @martinRenou does anything extra need to happen to update the conda-forge version?

PR #314, opened 2021-03-24
Convert frontend to typescript

ianhi:
1. Switch the frontend code from js to `ts`
2. Fix several small errors that typescript detected
3. Add proper `labextension:watch` script to package.json
4. Generally update the structure of the frontend to better track https://github.com/jupyter-widgets/widget-ts-cookiecutter for easier future maintainability.

I think the arguments for switching to typescript are:
1. Most of the jupyterlab ecosystem has a preference for typescript (including widgets)
2. Can catch errors (found some in this PR)
3. Can be more accessible to potential contributors coming from a pure python background. (anecdotal of course but this was very much true for me)

@martinRenou what do you think?

PR #264, opened 2020-09-08
remove unnecessary traitlets to fix savefig and display(fig) errors

ianhi:
fixes: https://github.com/matplotlib/ipympl/issues/252

I also preemptively removed several other attributes that aren't used by the `js` and so do not need to be traitlets.

I left `force_full`, `_current_image_mode` and `_dpi_ratio` as these have code of the js side that could use them, although currently all of the values for those are hardcoded in the js and the traitlets aren't used.

attn: @martinRenou @tacaswell 

Could a release be made after merging this?

PR #317, opened 2021-04-06
Handle new FigureCanvasBase.device_pixel_ratio.

QuLogic:
This is a corollary to
https://github.com/matplotlib/matplotlib/pull/19126.

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -86,8 +86,11 @@ class Toolbar(DOMWidget, NavigationToolbar2WebAgg):
         # Figure width in pixels
         pwidth = (self.canvas.figure.get_figwidth() *
                   self.canvas.figure.get_dpi())
-        # Scale size to match widget on HiPD monitors
-        width = pwidth / self.canvas._dpi_ratio
+        # Scale size to match widget on HiDPI monitors.
+        if hasattr(self.canvas, 'device_pixel_ratio'):  # Matplotlib 3.5+
+            width = pwidth / self.canvas.device_pixel_ratio
+        else:
+            width = pwidth / self.canvas._dpi_ratio
         data = "<img src='data:image/png;base64,{0}' width={1}/>"
         data = data.format(b64encode(buf.getvalue()).decode('utf-8'), width)
         display(HTML(data))
--- a/js/src/mpl_widget.js
+++ b/js/src/mpl_widget.js
@@ -72,6 +72,9 @@ export class MPLCanvasModel extends widgets.DOMWidgetModel {
     send_initialization_message() {
         if (this.ratio != 1) {
             this.send_message('set_dpi_ratio', { dpi_ratio: this.ratio });
+            this.send_message('set_device_pixel_ratio', {
+                device_pixel_ratio: this.ratio,
+            });
         }
 
         this.send_message('send_image_mode');


QuLogic:
There's a `_dpi_ratio` thing in `Canvas`, but I'm not sure if I need to do the same for `_device_pixel_ratio` (as set in `FigureCanvasBase` instead of `FigureCanvasWebAgg`).

ianhi:
Thanks!

This looks safe to merge prior to the the matplotlib PR correct? I think it makes sense to merge this prior to #314

QuLogic:
It should be safe yes, as the new message will just be logged as unknown in older Matplotlib.

ianhi:
> There's a _dpi_ratio thing in Canvas, but I'm not sure if I need to do the same for _device_pixel_ratio (as set in FigureCanvasBase instead of FigureCanvasWebAgg).

That's all the way back from the very first commit https://github.com/matplotlib/ipympl/commit/047ed9551836be1f3a803fd0bc5dce0e10b856be with a comment by   saying: `# Must declare the superclass private members.` @SylvainCorlay you left that comment do you know why they need to be declared? It's not clear to me why we need to transform those into traitlet variables if we aren't syncing them to the frontend.

ianhi:
Actually looking back at https://github.com/matplotlib/ipympl/pull/264 I'm not sure that that comment is still relevant. At the very least nothing bad happened when we removed all those traitlets. 

So in conclusion I don't think you need to change anything more. If anything we could even remove the `_dpi_ratio` being a traitlet.

ianhi:
Thanks!

SylvainCorlay:
Sorry I should have looked into this earlier.

One issue with fetching the device pixel ratio from the front-end (which is what we used to do in the past) is that it may depend on the browser window, so that different views of the same figure would have different values for that property.

ianhi:
> One issue with fetching the device pixel ratio from the front-end (which is what we used to do in the past) 

As far as I can tell that was how ipympl behaved prior to this PR as well. This PR just sends the same information to where it needs to be on the python side for mpl 3.5. 

Is there any way to deal with this? It seems to me that the frontend GUI framework is necessarily the source of truth for this information.

SylvainCorlay:
> As far as I can tell that was how ipympl behaved prior to this PR as well. This PR just sends the same information to where it needs to be on the python side for mpl 3.5.

I thought I removed that logic at some point because of that problem. If I recall correctly, we decided for an arbitrary fixed number (which could be overwritten in the backend).

> Is there any way to deal with this? It seems to me that the frontend GUI framework is necessarily the source of truth for this information.

Setting an arbitrary number may be good enough, since the ipympl front-end isn't really meant for print, pixel-units may be good enough.

ianhi:
> I thought I removed that logic at some point because of that problem. If I recall correctly, we decided for an arbitrary fixed number (which could be overwritten in the backend).

I think this is quantity we are talking about?

https://github.com/matplotlib/ipympl/blob/beec42bd7907e03cacfb1e58884307b24fbd3bb6/js/src/mpl_widget.js#L53


On my monitor at least (not hidpi) this is mostly affected by the zoom level of the jupyterlab tab.

SylvainCorlay:
What I meant is that it is impacted if you open the same notebook with two different computers at the same time. As we are working on live collaboration, this may happen in the near future.

SylvainCorlay:
Note that I am totally fine with keeping this for now, but we may want to keep this issue in mind. A solution could be to completely get rid of that variable.

PR #343, opened 2021-09-14
Implement image/png repr

martinRenou:
cc. @ianhi @SylvainCorlay @tacaswell 

This seems to properly work in JupyterLab, but we need to push a fix to ipywidgets to have it working in classic Jupyter Notebook as well.

ipympl_lab

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -161,6 +161,8 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
 
         self.on_msg(self._handle_message)
 
+        self._rendered = False
+
     def _handle_message(self, object, content, buffers):
         # Every content has a "type".
         if content['type'] == 'closing':
@@ -168,6 +170,11 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
         elif content['type'] == 'initialized':
             _, _, w, h = self.figure.bbox.bounds
             self.manager.resize(w, h)
+        elif content['type'] == 'ack':
+            update_display(
+                self._repr_mimebundle_(),
+                raw=True, display_id='matplotlib_{0}'.format(self._model_id)
+            )
         else:
             self.manager.handle_json(content)
 
@@ -197,39 +204,32 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
 
     def send_binary(self, data):
         self.send({'data': '{"type": "binary"}'}, buffers=[data])
-        update_display(self, display_id='matplotlib_{0}'.format(self.figure.number))
 
     def new_timer(self, *args, **kwargs):
         return TimerTornado(*args, **kwargs)
 
     def _repr_mimebundle_(self, **kwargs):
-        if self._view_name is not None:
-            # now happens before the actual display call.
-            self._handle_displayed(**kwargs)
-            plaintext = repr(self)
-            if len(plaintext) > 110:
-                plaintext = plaintext[:110] + '…'
-            # The 'application/vnd.jupyter.widget-view+json' mimetype has not been registered yet.
-            # See the registration process and naming convention at
-            # http://tools.ietf.org/html/rfc6838
-            # and the currently registered mimetypes at
-            # http://www.iana.org/assignments/media-types/media-types.xhtml.
+        # now happens before the actual display call.
+        self._handle_displayed(**kwargs)
+        plaintext = repr(self)
+        if len(plaintext) > 110:
+            plaintext = plaintext[:110] + '…'
 
-            buf = io.BytesIO()
-            self.figure.savefig(buf, format='png', dpi='figure')
-            data = "<img src='data:image/png;base64,{0}'/>"
-            data = data.format(b64encode(buf.getvalue()).decode('utf-8'))
+        buf = io.BytesIO()
+        self.figure.savefig(buf, format='png', dpi='figure')
+        data_url = b64encode(buf.getvalue()).decode('utf-8')
 
-            data = {
-                'text/plain': plaintext,
-                'text/html': data,
-                'application/vnd.jupyter.widget-view+json': {
-                    'version_major': 2,
-                    'version_minor': 0,
-                    'model_id': self._model_id
-                }
+        data = {
+            'text/plain': plaintext,
+            'image/png': data_url,
+            'application/vnd.jupyter.widget-view+json': {
+                'version_major': 2,
+                'version_minor': 0,
+                'model_id': self._model_id
             }
-            return data
+        }
+
+        return data
 
     def _ipython_display_(self, **kwargs):
         """Called when `IPython.display.display` is called on a widget.
@@ -240,9 +240,10 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
         if ipython_version_info >= (6, 1):
             raise NotImplementedError
 
-        data = self._repr_mimebundle_(**kwargs)
-        if data:
-            display(data, raw=True, display_id='matplotlib_{0}'.format(self.number))
+        display(
+            self._repr_mimebundle_(**kwargs),
+            raw=True, display_id='matplotlib_{0}'.format(self._model_id)
+        )
 
     if matplotlib.__version__ < '3.4':
         # backport the Python side changes to match the js changes
@@ -317,7 +318,10 @@ class FigureManager(FigureManagerWebAgg):
     def show(self):
         if self.canvas._closed:
             self.canvas._closed = False
-            display(self.canvas)
+            display(
+                self.canvas,
+                display_id='matplotlib_{0}'.format(self.canvas._model_id)
+            )
         else:
             self.canvas.draw_idle()
 
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -4,7 +4,8 @@ from base64 import b64encode
 import json
 import io
 
-from IPython.display import display, HTML
+from IPython.display import update_display, display, HTML
+from IPython import version_info as ipython_version_info
 
 from ipywidgets import DOMWidget, widget_serialization
 from traitlets import (
@@ -83,16 +84,8 @@ class Toolbar(DOMWidget, NavigationToolbar2WebAgg):
     def export(self):
         buf = io.BytesIO()
         self.canvas.figure.savefig(buf, format='png', dpi='figure')
-        # Figure width in pixels
-        pwidth = (self.canvas.figure.get_figwidth() *
-                  self.canvas.figure.get_dpi())
-        # Scale size to match widget on HiDPI monitors.
-        if hasattr(self.canvas, 'device_pixel_ratio'):  # Matplotlib 3.5+
-            width = pwidth / self.canvas.device_pixel_ratio
-        else:
-            width = pwidth / self.canvas._dpi_ratio
-        data = "<img src='data:image/png;base64,{0}' width={1}/>"
-        data = data.format(b64encode(buf.getvalue()).decode('utf-8'), width)
+        data = "<img src='data:image/png;base64,{0}'/>"
+        data = data.format(b64encode(buf.getvalue()).decode('utf-8'))
         display(HTML(data))
 
     @default('toolitems')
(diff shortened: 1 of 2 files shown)


js/src/mpl_widget.js
```diff
@@ -186,7 +188,11 @@ export class MPLCanvasModel extends widgets.DOMWidgetModel {
         this.image.src = image_url;
 
         // Tell Jupyter that the notebook contents must change.
-        this.send_message('ack');
+        if (!this.acknowledged_rendered) {
+            this.send_message('ack');
```
martinRenou, line 192: Why do we need a round-trip to the front-end??

martinRenou: I tried implementing it without the round-trip but I wasn't able. I need to look deeper into that.

martinRenou:
IDK what Github uses for rendering Notebooks, but it seems to correctly handle a mix of jupyter-widgets mimetype and `image/png` mimetype:

github_render

nbviewer is not able to render it

nbviewer

SylvainCorlay:
> nbviewer is not able to render it

Yes, that is expected as per our discussion on data type priority and the html renderer.

ianhi:
Does this depend on a specific vesion of jupyterlab? On the binder link for this PR I still get the widget mimebundle taking precedence:

Peek 2021-06-30 14-08

martinRenou:
@ianhi I was not aware of this binder link for PRs, is this really using a dev installation of ipympl? I see ipympl in the `environment.yml`: https://github.com/matplotlib/ipympl/blob/master/environment.yml#L8

martinRenou:
Relies on https://github.com/jupyter/notebook/pull/6181 and https://github.com/jupyter-widgets/ipywidgets/pull/3280 for Jupyter Notebook

ianhi:
> @ianhi I was not aware of this binder link for PRs, is this really using a dev installation of ipympl? I see ipympl in the environment.yml: https://github.com/matplotlib/ipympl/blob/master/environment.yml#L8

:/ indeed it seems to be wrong. Ideally the bot that commented gives you a link to binder based on the PR in question.

ianhi:
> Does this depend on a specific vesion of jupyterlab?

Testing locally it seems that this doesn't work on jlab 3.0.14 but does on 3.1.12 so something in rendering priorities or response to errors changed between 3.0 and 3.1?

--- 

potential problem:

If I understand correctly this approach works because jlab falls back on the lower priority mimetypes when there is an error loading the widget state. This error shows up because the full widget state wasn't saved, but if you check the `Save widget state automatically` option then the png is never showed and the frontend stalls out waiting for a draw message (it doesn't render until there is some back and forth with pyton) even though I think it should have all the information it needs to draw itself.

Peek 2021-09-15 12-15

(Somewhere I've written down why I think this is happening and how it might be fixed, I'll try to find that)

martinRenou:
> This error shows up because the full widget state wasn't saved, but if you check the Save widget state automatically option then the png is never showed

Yes, that should be the intended behavior. We could fix this by including the png data as part of the widget model.

martinRenou:
nbviewer and nbconvert should take this into account if this gets merged: https://github.com/jupyter/nbconvert/pull/1643

ipympl/backend_nbagg.py
```diff
@@ -208,6 +208,44 @@ def send_binary(self, data):
...
+        if hasattr(self, '_handle_displayed'):
+            self._handle_displayed(**kwargs)
+        plaintext = repr(self)
+        if len(plaintext) > 110:
+            plaintext = plaintext[:110] + '…'
+
+        buf = io.BytesIO()
+        self.figure.savefig(buf, format='png', dpi='figure')
+        data_url = b64encode(buf.getvalue()).decode('utf-8')
+
```
martinRenou, line 222: @ianhi it seems that at this point in the code we still don't have the figure rendered (at least not the first time `display` gets called, so we need to call `update_display` later). It would be nice to not have to use `update_display` though, as it makes the widget flicker, and it wouldn't work properly from scripting tools like `nbconvert --execute`.

I've tried forcing rendering the figure there using `self.draw()` or `self.figure.draw(self.get_renderer())` here, but it doesn't work. Do you know how I could force the figure to be rendered **before** having received the `initialized` message from the front-end?

ianhi: In what way is it not working? Is the draw not happening or is savefig giving the wrong buffer or something?

Two potential avenues:
1. Maybe some of the trickery i proposed here: https://github.com/matplotlib/ipympl/issues/290#issuecomment-755377055
    - I think that the core of #290 is about when draws happen and the issue is back and forth with the frontend about `dpi` scaling.  
2. Can you try drawing and then instead of getting the buffer from savefig getting it similar to how it is done here: https://github.com/matplotlib/matplotli/blob/88ba86bcd41991fa81372b2f8c44710348195035/lib/matplotlib/backends/backend_webagg_core.py#L194-L196

martinRenou: > In what way is it not working? 

The figure seems to have the right size but the buffer is completely empty, resulting in a white png.

> Can you try drawing and then instead of getting the buffer from savefig getting it similar to how it is done here:

IIRC I tried doing this already, with `renderer.buffer_rgba` but it was giving the same result. I can try again though.

> when draws happen and the issue is back and forth with the frontend about dpi scaling

That might be the issue...

ianhi:
ohhh one more thing about this approach. I think we will need to call update_display on every draw call (or more parsimoniously whenever the notebook is saved) because otherwise you don't capture any updates to the figure that happen after the initial display call (panning, zooming, adding elements, manipulating with sliders, etc)

Peek 2021-09-16 11-31


Can widgets listen to notebook save events? I think updating the embedded png on saving is more reasonable than on every draw in order to avoid the flickering you described.

martinRenou:
> ohhh one more thing about this approach. I think we will need to call update_display on every draw call (or more parsimoniously whenever the notebook is saved) because otherwise you don't capture any updates to the figure that happen after the initial display call (panning, zooming, adding elements, manipulating with sliders, etc)

Yeah. Though `update_display` makes the widget flicker and it is quite annoying. We might be able to improve JupyterLab/Notebook to not have this flickering though.
Also this means sending the data twice for each update, once for the widget, and once for the `image/png` repr, so not really ideal.

> Can widgets listen to notebook save events? I think updating the embedded png on saving is more reasonable than on every draw in order to avoid the flickering you described.

Unfortunately not on the Python side. Technically, the kernel doesn't even know what a Notebook is.

I'm tempted to say we could ignore this particular issue for now. As what we have is already quite a big improvement. And other Matplotlib backends are also not able to save the latest state of the plot without an `update_display` call.

ipympl/backend_nbagg.py
martinRenou: Actually this is expected to be blank...

When running the following code:
```python
fig = plt.figure()
plt.plot(np.sin(np.linspace(0, 20, 100)));
```
The first line will trigger a display of the widget even though we don't have any plot yet (which results in a blank mimetype). In the current state of the PR, the second line will trigger an `update_display`.

So I believe we can't get rid of `update_display` in the current state...

martinRenou:
Now using a `post_execute` cell hook (similar to what the inline backend does), this allows getting rid of the roundtrip to the client and not having to use the `update_display` (removing the flickering).

This seems to work nicely, though there is still a bit of polishing to do.

martinRenou:
cc. @ianhi @SylvainCorlay @tacaswell

This seems to work nicely in JupyterLab now! It makes ipympl closer to the inline backend in behavior and implementation.

- It will work in classic Jupyter Notebook when `notebook` gets released and `widgetsnbextension` gets released with https://github.com/jupyter-widgets/ipywidgets/pull/3280.
- It will work with `nbconvert` when https://github.com/jupyter/nbconvert/pull/1643 is released

We can merge this PR now and not wait for `notebook` and `nbconvert` to update.

martinRenou:
ipympl

SylvainCorlay:
Let's get this in as it is a net improvement over the current situation, and tag a patch release of ipympl.

martinRenou:
Let's make a minor release instead, as the API changes a tiny bit: it's not `plt.figure()` which makes the display anymore, so wrapping it inside an `Output` widget doesn't work anymore.

SylvainCorlay:
Placing show in an output widget shoulds still work as expected though.

jklymak:
Thanks for the persistence on this one!  Sounds like it was tricky...

ipympl/backend_nbagg.py
```diff
@@ -83,16 +83,8 @@ def __init__(self, canvas, *args, **kwargs):
...
-        # Figure width in pixels
-        pwidth = (self.canvas.figure.get_figwidth() *
-                  self.canvas.figure.get_dpi())
-        # Scale size to match widget on HiDPI monitors.
-        if hasattr(self.canvas, 'device_pixel_ratio'):  # Matplotlib 3.5+
-            width = pwidth / self.canvas.device_pixel_ratio
-        else:
-            width = pwidth / self.canvas._dpi_ratio
-        data = "<img src='data:image/png;base64,{0}' width={1}/>"
-        data = data.format(b64encode(buf.getvalue()).decode('utf-8'), width)
```
ianhi, line 95: @martinRenou why is it ok to remove this dpi scaling? It seems like we will now have imperfect behavior on hi dpi displays.

ianhi:
> Let's make a minor release instead, as the API changes a tiny bit: it's not plt.figure() which makes the display anymore, so wrapping it inside an Output widget doesn't work anymore.

I can see this making #290 harder because now figures aren't shown until the end of cell execution? Or maybe that can be made to work with an explicit call to `plt.show` in a notebook cell?

ipympl/backend_nbagg.py
martinRenou: This might have been a mistake, we should open an issue to track this

jklymak: I'll do it, actually.  If I understand correctly, the PNG doesn't get saved at high res.

SylvainCorlay:
> I can see this making #290 harder because now figures aren't shown until the end of cell execution? Or maybe that can be made to work with an explicit call to plt.show in a notebook cell?

Yes, I actually think that the interactive mode (`plt.ion()`) is somewhat intrinsically broken. I would prefer it to be `off` by default and explicit calls to `plt.show()` to be required...

ipympl/backend_nbagg.py
SylvainCorlay: Note: we should not determine the dpi based on the browser because it could differ between clients.
It should be a fixed value.

nvaytet:
Hi, unfortunately, I am still seeing this error after re-opening the notebook.
Screenshot at 2022-06-29 16-31-42

The console log from Chrome:
Screenshot at 2022-06-29 16-33-54

Versions:
```
jupyter_client            7.3.1              pyhd8ed1ab_0    conda-forge
jupyter_core              4.10.0           py39hf3d152e_0    conda-forge
jupyter_server            1.17.0             pyhd8ed1ab_0    conda-forge
jupyterlab                3.4.2              pyhd8ed1ab_0    conda-forge
jupyterlab_pygments       0.2.2              pyhd8ed1ab_0    conda-forge
jupyterlab_server         2.14.0             pyhd8ed1ab_0    conda-forge
jupyterlab_widgets        1.1.0              pyhd8ed1ab_0    conda-forge

matplotlib                3.5.2            py39hf3d152e_0    conda-forge
matplotlib-base           3.5.2            py39h700656a_0    conda-forge
matplotlib-inline         0.1.3              pyhd8ed1ab_0    conda-forge

ipydatawidgets            4.3.1.post1        pyhc268e32_0    conda-forge
ipykernel                 6.13.0           py39hef51801_0    conda-forge
ipympl                    0.9.1              pyhd8ed1ab_0    conda-forge
ipython                   8.4.0            py39hf3d152e_0    conda-forge
ipython_genutils          0.2.0                      py_1    conda-forge
ipywidgets                7.7.0              pyhd8ed1ab_0    conda-forge
```

Thanks for any help!

jklymak:
@nvayet, I'd open a new issue - but for what it is worth, I have the same problem as you....

PR #344, opened 2021-09-15
update binder env

ianhi:
Trying to get the binder bot that comments on PRs to install the branch rather than version `0.6.3`

Noticed in https://github.com/matplotlib/ipympl/pull/343

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,6 +3,5 @@ channels:
   - conda-forge
 dependencies:
   - numpy
-  - jupyterlab=3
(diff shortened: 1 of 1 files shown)


ianhi:
not yet working. will try to borrow from the jlab-vim setup: https://github.com/jupyterlab-contrib/jupyterlab-vim/tree/master/binder

martinRenou:
I guess we need a `postBuild` file with `pip install -e .`

martinRenou:
Follow up in #344

PR #357, opened 2021-09-22
Binder dev install

martinRenou:
Take over https://github.com/matplotlib/ipympl/pull/344

I cherry-picked your commit @ianhi so you still get attribution


martinRenou:
It seems to work as intended now :)

binder

PR #358, opened 2021-09-22
Install ipympl from conda

SylvainCorlay:
Adopting the same patter of reseting `stable` on `master` when the new version of the conda package is available.

cc @martinRenou


SylvainCorlay:
Arg, this will break the binder on PRs...

martinRenou:
Yeah. I was actually wondering if it was possible to have both.

ianhi:
We could try to have a github action update the `stable` branch whenever a new release is made and then have the  binder link in the readme point to that, but have the master binder install as it currently does.

PR #369, opened 2021-10-07
Make the plot image part of the widget model

martinRenou:
widget_state

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -4,10 +4,22 @@ import io
 import json
 from base64 import b64encode
 
-from PIL import Image
-
+import matplotlib
 import numpy as np
-
+from IPython import get_ipython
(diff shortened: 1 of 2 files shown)


ianhi:
> Giving this more thought, I don't know of a way of doing it actually. I believe ipywidgets will always sync properties. We might not have a similar option to the sync=False in Python.

> EDIT: Though we might be able to overwrite the sync method https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/base/src/widget.ts#L393 to prevent the sync of the _image attribute.


Actually I think that the frontend never synces data back unless you explicitly call `model.save_changes` as in https://github.com/cytoscape/ipycytoscape/blob/5e9695df32f00d47d71b053edf3bccbac3388b68/src/widget.ts#L363-L364 so since we never call that maybe we don't even need to override sync?

PR #376, opened 2021-10-08
Make image as part of the widget model - alternate implementation

martinRenou:
cc. @ianhi

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -4,10 +4,18 @@ import io
 import json
 from base64 import b64encode
 
+try:
+    from collections.abc import Iterable
+except ImportError:
+    # Python 2.7
+    from collections import Iterable
+
 import matplotlib
+import numpy as np
 from IPython import get_ipython
 from IPython import version_info as ipython_version_info
 from IPython.display import HTML, display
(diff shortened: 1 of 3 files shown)


martinRenou:
Though I believe this doesn't work properly (yet?). It does not work if there is no front-end. For example it doesn't work if you execute the Notebook with nbconvert.

martinRenou:
I would be in favor of #369 actually. Ideally, we would have it as a synced trait, but instead of sending the entire image every single time, we would sync the backend/frontend traits sending only image diffs.

This is an optimization that I would like to see being implemented more generally in ipywidgets. For example, when syncing a List trait it would be better to send diffs of this list instead of the entire list.

ianhi:
> I would be in favor of #369 actually. Ideally, we would have it as a synced trait, but instead of sending the entire image every single time, we would sync the backend/frontend traits sending only image diffs.

That sounds good to me. The big reason I preferred this approach was the effect on performance and not painting ourselves into a corner w.r.t true blitting.


> This is an optimization that I would like to see being implemented more generally in ipywidgets. For example, when syncing a List trait it would be better to send diffs of this list instead of the entire list.

Definitely agree!

Issue #299, opened 2021-02-01
Feature req: Show toolbar on mouse-over


bdch1234:
Would it be possible to add functionality for hiding the toolbar when mouse moves away, and show when in the canvas, like corresponding functionality in plotly?

PR #396, opened 2021-11-12
Refactor toolbar

martinRenou:
Fix https://github.com/matplotlib/ipympl/issues/299

https://user-images.githubusercontent.com/21197331/141501289-093bc413-8a3d-4ca9-99e4-13b4c08f2742.mp4


martinRenou:
There are some CSS issues still

martinRenou:
https://user-images.githubusercontent.com/21197331/142438318-666927a6-d0c9-4b98-8b03-45d6432a719d.mp4

martinRenou:
@tacaswell @SylvainCorlay @ianhi if you have any thoughts :) Feedback would be super welcome

martinRenou:
The plot is now centered in the cell output

ianhi:
> @tacaswell @SylvainCorlay @ianhi if you have any thoughts :) Feedback would be super welcome

Will look tonight or this weekend

ianhi (changes requested):
Sorry for the delay @martinRenou 

**general refactor**
I am :+1:  on the code simplification

**fade in/out**
:+1: 
I think the fade in and out is nice and my one suggestion is that it should be toggleable by a user.


**plot centering**

Currently :-1: but open to being convinced or improving this.


Was this necessary to get the fading to work?

The two main issues that the centering brought for me are:

1. Weird gap to the toolbar when the plot is small. This makes it inconvenient to use the toolbar because you have to move your mouse so far
image


2. The centering has a big impact on the dynamics of resizing the plot. It doubles the rate of resizing horizontally while leaving vertical resizing the same speed which feels very un-intuitive:
toolbar-center-resize


I think what happened is that this breaks the assumption that when were resizing a window the opposite edges (top and left) will stay in their positions. Maybe this could be reconfigured so that those edges stay fixed while resizing? Otherwise I think this ends up being more detrimental than beneficial (users can already center the plot using ipywidgets although that isn't super smooth).

martinRenou:
Thanks for the review @ianhi :)

I can revert the centering of the plot. This was not the initial purpose of the PR anyway.

martinRenou:
@ianhi Thanks again for your review. I took your comments into account:

- Reverted the centering of the plot
- Made the fade-in fade-out effect optional (enabled by default)

css/mpl_widget.css
```diff
@@ -18,9 +24,12 @@
 /* Figure */
 
 .jupyter-matplotlib-figure {
-    width: auto;
+    width: fit-content;
     height: auto;
     overflow: hidden;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
```
ianhi, line 32: @martinRenou i thought the centering was reverted?

martinRenou: It was.. This looks like a leftover. But this does not seem to affect the plot? It is not centered when I try latest.

ianhi: huh i was just seeing this behavior. let me try in a fresh env

ianhi: I see this on the binder link - maybe this is browser based. I'm on Firefox 93

ianhi: indeed on chrome it is not centered, but it is on firefox:

left = chrome
right = firefox
Peek 2022-01-19 18-15

martinRenou: I wish the visual regression tests caught that. We might be able to use playwright with Firefox as well. #422

martinRenou: Thanks a lot for catching this!

PR #398, opened 2021-11-15
slow down pre-commit autoupdates

ianhi:
The default is `weekly` which I don't think is really necessary and introduces a decent amount of PR noise.


martinRenou:
Thank you! :P

PR #406, opened 2021-12-14
Refactor image embedding logic

martinRenou:
Refactor the image embedding logic.

This fixes an issue in the case of `nbconvert --to html` that was preventing the plot image to render.

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -51,6 +51,7 @@ from traitlets import (
     Float,
     Instance,
     List,
+    Tuple,
     Unicode,
     default,
 )
@@ -184,8 +185,7 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
     # This will still be used by ipywidgets in the case of embedding.
     _data_url = Any(None).tag(sync=True)
 
-    _width = CInt().tag(sync=True)
-    _height = CInt().tag(sync=True)
+    _size = Tuple([0, 0]).tag(sync=True)
 
     _figure_label = Unicode('Figure').tag(sync=True)
     _message = Unicode().tag(sync=True)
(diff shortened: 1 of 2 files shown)


ipympl/backend_nbagg.py
```diff
@@ -306,28 +308,36 @@ def _repr_mimebundle_(self, **kwargs):
...
                 </div>
-                <img src='data:image/png;base64,{}' width={}/>
+                <img src='{}' width={}/>
             </div>
         """.format(
             self._figure_label, self._data_url, width
         )
 
+        # Update the widget model properly for HTML embedding
+        self._size = (width, height)
```
martinRenou, line 336: This is one important change here.

We need to make the plot size part of the widget model in the case of `nbconvert --to html` otherwise there is no registered size and the canvas would be 0 sized.

src/mpl_widget.ts
```diff
@@ -275,17 +287,49 @@ export class MPLCanvasModel extends DOMWidgetModel {
         this.image = new Image();
 
         this.image.onload = () => {
+            // In case of an embedded widget, the initial size is not correct
+            // and we are not receiving any resize event from the server
+            if (this.disabled) {
```
martinRenou, line 292: This is the second important change in this PR.

Here, we need a special case for when the widget is disabled (no counterpart in the kernel, or no kernel at all), where we display the image that is part of the model.

ipympl/backend_nbagg.py
```diff
@@ -184,8 +185,7 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
     # This will still be used by ipywidgets in the case of embedding.
     _data_url = Any(None).tag(sync=True)
 
-    _width = CInt().tag(sync=True)
-    _height = CInt().tag(sync=True)
+    _size = Tuple([0, 0]).tag(sync=True)
```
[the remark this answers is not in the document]
ianhi: I like it

src/mpl_widget.ts
ianhi: Am I right that this will come up in many situations, not just nbconvert, right?

ianhi:
Haven't tried it out, but this all looks pretty sensible to me.

martinRenou:
Thanks for looking into it!

I will merge and make a patch release with this soonish.

PR #407, opened 2021-12-15
Improve toolbar positioning

martinRenou:
Using percentages for the positioning of the toolbar worked fine for small plots, but it didn't look great for big screens


martinRenou:
Failures are unrelated, the npm registy seems down

Issue #409, opened 2021-12-15
Replace the ipykernel dependency with an IPython dependency?


martinRenou:
## Describe the issue

In the `setup.py` file and in the conda-forge recipe, there is an ipykernel dependency that we could maybe get rid of. We should probably reduce the dependency to an IPython dependency, which we use directly in the codebase.

martinRenou:
The conda-forge bot is complaining about the following:

```
Missing dependencies:
- traitlets
- jupyter-packaging
- numpy
- ipython
- pillow
- ipython_genutils

Unused dependencies:
- six
- ipykernel
```

stonebig:
is it still mandatory to have the contraint ipython<8 ?

PR #418, opened 2022-01-15
allow  ipython-8.0

stonebig:
now that ipython-8 is there, maybe it's not a problem https://github.com/matplotlib/ipympl/issues/409


martinRenou:
Thanks!

PR #421, opened 2022-01-19
Remove centering of plot on firefox

ianhi:
Followup to: afdebbfade53785434b7ac7fe3965480895b6ccd

which didn't have centering for chrome, but did for firefox

cc @martinRenou  this shouldn't do anything bad (I think?) and fixes the centering for me on firefox


martinRenou:
Thanks a lot! I'll do a release right away

martinRenou:
The visual regression tests were not passing I had to update the references. It looks like this was having an impact on the header position on Chrome? It is unclear to me why it was behaving differently on Firefox vs Chrome.

martinRenou:
The remaining failures are unrelated to this PR. I believe we need to do something similar as in bqplot https://github.com/bqplot/bqplot/pull/1440

Issue #426, opened 2022-02-07
Tests are failing on latest matplotlib


ianhi:
## Describe the issue

For example see: https://github.com/matplotlib/ipympl/actions/runs/1806282769

Getting a lot of these errors so something must have changed upstream w.r.t colorbars i suppose

```

%matplotlib widget

import matplotlib.pyplot as plt
import numpy as np

fig = plt.figure()
plt.plot(np.sin(np.linspace(0, 20, 100)));

Traceback:

---------------------------------------------------------------------------
TraitError                                Traceback (most recent call last)
Input In [1], in <module>
      3 import matplotlib.pyplot as plt
      4 import numpy as np
----> 6 fig = plt.figure()
      7 plt.plot(np.sin(np.linspace(0, 20, 100)))

File /usr/share/miniconda/envs/test/lib/python3.10/site-packages/matplotlib/pyplot.py:792, in figure(num, figsize, dpi, facecolor, edgecolor, frameon, FigureClass, clear, **kwargs)
    783 if len(allnums) == max_open_warning >= 1:
    784     _api.warn_external(
    785         f"More than {max_open_warning} figures have been opened. "
    786         f"Figures created through the pyplot interface "
   (...)
    789         f"warning, see the rcParam `figure.max_open_warning`).",
    790         RuntimeWarning)
--> 792 manager = new_figure_manager(
    793     num, figsize=figsize, dpi=dpi,
    794     facecolor=facecolor, edgecolor=edgecolor, frameon=frameon,
    795     FigureClass=FigureClass, **kwargs)
    796 fig = manager.canvas.figure
   2893     # This cursor will be set after the initial draw.

File /usr/share/miniconda/envs/test/lib/python3.10/site-packages/traitlets/traitlets.py:606, in TraitType.__set__(self, obj, value)
    604     raise TraitError('The "%s" trait is read-only.' % self.name)
    605 else:
--> 606     self.set(obj, value)

File /usr/share/miniconda/envs/test/lib/python3.10/site-packages/traitlets/traitlets.py:580, in TraitType.set(self, obj, value)
    579 def set(self, obj, value):
--> 580     new_value = self._validate(obj, value)
    581     try:
    582         old_value = obj._trait_values[self.name]

File /usr/share/miniconda/envs/test/lib/python3.10/site-packages/traitlets/traitlets.py:612, in TraitType._validate(self, obj, value)
    610     return value
    611 if hasattr(self, 'validate'):
[...]

PR #428, opened 2022-02-12
fix toolbar for mpl>3.6

ianhi:
Will fix: https://github.com/matplotlib/ipympl/issues/426 once https://github.com/matplotlib/matplotlib/pull/22454 goes in


ianhi:
@martinRenou I was about to bump versions and make a new release but realized I don't actually know what all needs to be bumped. In particular what do we do with `js_semver`?

https://github.com/matplotlib/ipympl/blob/fc0e08c7a88fd08eba6faa71660718a66856c9c6/ipympl/_version.py#L3

martinRenou:
Thanks! 

You can probably just make a Python package release. We didn't change the JS code since the last release.

Issue #411, opened 2021-12-21
Getting back toolbar on left (if set to top position) in 0.8.4?


sdbbs:
## Describe the issue

In an earlier version of ipympl (unfortunately, haven't recorded which), if I chose the positioning of the toolbar on top, it was automatically placed on the left; I have a partial record of that here - https://stackoverflow.com/questions/69106967/jupyter-matplotlib-widget-place-toolbar-horizontally-over-plot

However, I made an update recently, and it turns out, the hacks that I have documented previously, do not work anymore - also because now the default for `fig.canvas.toolbar_visible` seems to be `fade-in-fade-out` (apparently, when I documented above hacks, the default was `visible`):

So now, if you run the last code snippet in the answer of the linked SO post above (only thing I added was version printout):

```python
%matplotlib widget
import ipympl; print("ipympl version {}".format(ipympl.__version__))

import matplotlib.pyplot as plt
from ipywidgets import widgets, Layout
from IPython.display import display, HTML

out_widget_plot = widgets.Output()
with out_widget_plot:
    with plt.style.context("default"): # ggplot
        fig, (ax1) = plt.subplots(1, 1)

fig.suptitle('test title', fontsize=10) 
tline = ax1.plot([0,1,2], [10,20,30])
fig.tight_layout()
#fig.canvas.toolbar.button_style = 'info' # test: ok, works (but don't need it here)
#fig.canvas.toolbar.orientation = 'horizontal' # no error, but no effect either
fig.canvas.toolbar_position = 'top' # this also switches the orientation to horizontal, too
fig.canvas.header_visible = False # Gets rid of "Figure 1" on top

# the below CSS style override in Jupyter, allows the div that hosts the Matplotlib toolbar, to be placed "on top" of the div that hosts the plot:
# widget-label does the same, but for the bottom "x/y" indicator (which is why its position needs to also be controlled with margin)
sthtml = """
<style>
.jupyter-matplotlib-toolbar {
    overflow: visible;
    position: absolute;
    z-index: 1000;
}
.widget-label {
    color: var(--jp-widgets-label-color);
    font-size: var(--jp-widgets-font-size);
    overflow: hidden;
    text-overflow: ellipsis;
    white-space: nowrap;
    line-height: var(--jp-widgets-inline-height);
    z-index: 1000;
    margin-top: -41px;
    margin-left: 58px;
}
</style>
"""

[...]

PR #434, opened 2022-02-25
Create docs

ianhi:
Scattered over a few issues there has been mention of the need for a documentation page. This is a first pass at creating one. I think there are lots of great things we can add (e.g. css tricks https://github.com/matplotlib/ipympl/issues/411#issuecomment-998583424) but getting something out there is more important than getting everything in there. I think it's important for users to be able to go to a website to get install instructions rather than a github repo as I suspect that this is lends a sense of stability and officiallness to the project. It also makes it easier to have much more robust sections on installing and contributing. 

I think the big question I have is probably primarily for @tacaswell (and also @martinRenou ) which is where should this be hosted? I know how to set things up on readthedocs and so can can easily do that. But this is also a matplotlib subproject (unclear how official that makes things) and does that mean that this shoudl be hosted on matplotlib.org/ipympl? I think that would be my preferred URL if possible. 

Other thoughts:

1. One day my dream is to have the docs be immediately enabled as interactive via pyodide @martinRenou do you know how far we are from being able to do something like that?
2. We should have a more comprehensive section on using ipympl with ipywidgets
    - Thoughts on including mpl-interactions there? I'm biased but I find it to be the best way to pair ipywidgets with matplotlib.


docs/installing.md
```diff
@@ -0,0 +1,61 @@
...
+```bash
+conda install -c conda-forge nodejs
+jupyter labextension install @jupyter-widgets/jupyterlab-manager jupyter-matplotlib
+```
+
+
+<!--
+## Sagemath CoCalc
+TODO
+-->
```
ianhi, line 28: Anyone have an idea if/how to get ipympl working in sagemath/cocalc?

martinRenou:
> One day my dream is to have the docs be immediately enabled as interactive via pyodide @martinRenou do you know how far we are from being able to do something like that?

We're pretty close actually :) 

I'd really like to have a simple [replite](https://github.com/jtpio/replite/) console at the beginning of the docs. But this requires widgets to work in JupyterLab console https://github.com/jupyter-widgets/ipywidgets/pull/3004

martinRenou:
I'm making some experiments on jupyterlite-sphinx: 

https://github.com/martinRenou/jupyterlite-sphinx
https://jupyterlite-sphinx.readthedocs.io/en/latest

ianhi:
bump @tacaswell

tacaswell:
I'm sorry, I wrote message here (and even checked it off my to-do list) but apparently never hit "send" 🐑 🤦🏻 

I think be best path to hosting is to use something like https://github.com/matplotlib/mpl-gui/blob/main/.github/workflows/docs.yml to put the built docs on github.  We just this weekend moved to self-hosting the docs on digital ocean (still behind cloudflare as a CDN).  We will need to add a web-hook to notify that system when new docs are pushed.

attn @QuLogic

docs/installing.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Google Colab
+
+To use `ipympl` in colab follow the instructions [here](https://colab.research.google.com/github/googlecolab/colab-cdn-widget-manager/blob/main/notebooks/ipympl.ipynb)
+
```
ianhi, line 33: TODO:

add:

```python
from google.colab import output
output.enable_custom_widget_manager()
```

does that work? I guess the linked notebook here is out of date?

martinRenou: > I guess the linked notebook here is out of date?

It looks like it's outdated indeed. There have been multiple changes in the Colab API before it was working properly.

I believe it works nicely now with the code snippet I shared :) At least it worked like a charm the other day when I tried ipycanvas (which is similar in some ways to ipympl)

ianhi: done :)

ianhi:
Ok. I think I'm happy with this now. Can someone give a read over to review?

QuLogic:
> > We will need to add a web-hook to notify that system when new docs are pushed.
> 
> @QuLogic any chance of this (if not already done?) I don't think there's much left to do here

It will only make sense to do this after you have a branch with built docs in it.

ianhi:
> Ok. I think I'm happy with this now. Can someone give a read over to review?

Absent any objections I'm going to merge this. we can always iterate more later

ianhi:
@QuLogic we now have a `gh-pages` branch that contains built docs

docs/installing.md
ianhi: Now that there's less noise here: @williamstein I couldn't find an obvious way to do this after some brief searching and the one attempt I had didn't work. Would love to include instructions though

ianhi:
@QuLogic bump - this will eventually end up at https://matplotlib.org/ipympl correct?

tacaswell:
https://github.com/matplotlib/matplotlib.org/pull/15 is step 1

QuLogic:
The secret is just a random set of characters.

Issue #138, opened 2019-09-18
Save button should save in rcparams format, not "png"


thomasaarholt:
@tfblum noticed that the save button always saves images in the png format. In `plt.rcParams['savefig.format']`, one can specify the default saving format of `plt.savefig`, and it would be nice if the save button respected that.

That said, I think maybe the on-screen generated image already *is* a png? There are two references to "png" in the codebase:
[nbagg-backend](https://github.com/matplotlib/jupyter-matplotlib/blob/b281960d3d11d3f2190ed3da0a1e9844add344f1/ipympl/backend_nbagg.py#L106), [`handle_save`](https://github.com/matplotlib/jupyter-matplotlib/blob/8b1536618ac12427373ef90d041013024150891c/js/src/mpl_widget.js#L247) and [`handle_binary`](https://github.com/matplotlib/jupyter-matplotlib/blob/8b1536618ac12427373ef90d041013024150891c/js/src/mpl_widget.js#L318)

ianhi:
From #234 it's also seems that other savefig rcparams are not going to be respected. I guess the reason figure saving is structured the way it is right now is because we need javascript in order to open a download prompt. I see two potential solutions:

1. Button pressed -> open jupyter dialog for savefig directory -> plt.savefig  

2. Use js to download the file but use plt.savefig to generate the buffer first.
    1. **js** On save button being pressed send a message to python side
    2. **py** `buf = Bytes(); plt.savefig(buf')`
    3. **py** `self.send({'name':'saving-figure'}, [buf])`
    4. **js** on:msg use current saving infrastructure.

jkochNU:
Hi - running into the same limitation. Are there any concrete plans to move one of the suggested solutions forward?

ianhi:
@jkochNU I think this issue is still looking for a champion to push it forward. Could be you :)

If anyone is interested in working on this feel free to ask questions here and i can try to help

jkochNU:
@ianhi Happy to try, though this may turn out to be above my level of competence. I can navigate the py side, but am less familiar with the js side. Which js file is the one involved in (i) and (iv)?

ianhi:
@jkochNU  awesome! Dev installation instructions here: https://github.com/matplotlib/ipympl#for-a-development-installation-requires-nodejs

 Here are some more concrete steps to solve this
Saving starts when a user clicks the download button:  
### Python:

[...]

PR #443, opened 2022-03-27
button for canvas toolbar which enables creating a pdf of the figure

jkochNU:
As discussed with @ianhi, this is an attempt to (partially) address #138. 
This would add an additional button to the toolbar associated with a `figure.canvas`. Once clicked, a pdf is generated with `savefig` and then handed over to the frontend inside a new tab. 

image

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -147,11 +147,13 @@ class Toolbar(DOMWidget, NavigationToolbar2WebAgg):
             'move': 'arrows',
             'download': 'floppy-o',
             'export': 'file-picture-o',
+            'make_pdf': 'file-pdf'
         }
 
         download_item = ('Download', 'Download plot', 'download', 'save_figure')
+        make_pdf = ('Create pdf', 'Create pdf from figure', 'make_pdf', 'makepdf')
 
-        toolitems = NavigationToolbar2.toolitems + (download_item,)
+        toolitems = NavigationToolbar2.toolitems + (download_item,) + (make_pdf,)
 
         return [
             (text, tooltip, icons[icon_name], method_name)
@@ -274,6 +276,9 @@ class Canvas(DOMWidget, FigureCanvasWebAggCore):
         elif content['type'] == 'set_dpi_ratio':
             Canvas.current_dpi_ratio = content['dpi_ratio']
             self.manager.handle_json(content)
+
+        elif content['type'] == 'toolbar_button' and content['name'] == 'makepdf':
+            self._send_savefig_pdf()
 
         else:
             self.manager.handle_json(content)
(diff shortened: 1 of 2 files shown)


ipympl/backend_nbagg.py
```diff
@@ -450,6 +455,11 @@ def handle_key(key):
 
         handle_key_press = handle_key_release = _handle_key
 
+    def _send_savefig_pdf(self):
+        buf = io.BytesIO()
+        self.figure.savefig(buf, format='pdf', dpi='figure')
+        self.send({'data': '{"type": "makepdf"}'}, buffers=[buf.getbuffer()])
```
ianhi, line 461: Changing this a little bit will allow generalizing to more than just PDFs.

1. Remove the arguments to savefig so that it follows the rcparams
2. check the rcparams format and use that to construct the mimetype
3. send that mimetype info along to the frontend

```suggestion
        self.figure.savefig(buf)
        format = matplotlib.rcParams['savefig.format']
        # some code intepreting the mimetype
        # e..g application/pdf, image/svg+xml, image/png etc
        self.send({'data': {"type": "download-fig", "format":format, "mimetype":mimetype}}, buffers=[buf.getbuffer()])
```

src/mpl_widget.ts
```diff
@@ -157,6 +157,14 @@ export class MPLCanvasModel extends DOMWidgetModel {
         document.body.removeChild(save);
     }
 
+    handle_makepdf(msg: any, dataviews: any) {
+        const url_creator = window.URL || window.webkitURL;
+        const buffer = new Uint8Array(dataviews[0].buffer);
+        const blob = new Blob([buffer], { type: 'application/pdf' });
```
ianhi, line 163: With other comment extract the mimetype from the msg

src/mpl_widget.ts
```diff
@@ -157,6 +157,14 @@ export class MPLCanvasModel extends DOMWidgetModel {
...
+        const image_url = url_creator.createObjectURL(blob);
+        window.open(image_url, '_blank');
```
ianhi, line 165: I'm pretty sure  that we need to revoke the object url after downloading

ianhi (changes requested):
This is a good start! Glad that you managed all the interconnections between python and typescript.

That said I don't think that the correct solution is to limit ourselves to pdfs here. Instead we could have the button mean "Save according to rcparams" and a second button (probably with a different icon) that would do the current behavior of downloading. 

For now just adding a new button for saving with rcparams is a good step I think. inline comments should show how to accomplish that.

ianhi:
@jkochNU heads up that the pre-commit bot push a commit with styling changes. You can either `git pull` or on your next push do `git push -f`to override that commit (but then it will happen again to auto style things). I'd also recommend installing it locally with `pre-commit install`

jkochNU:
> That said I don't think that the correct solution is to limit ourselves to pdfs here. Instead we could have the button mean "Save according to rcparams" and a second button (probably with a different icon) that would do the current behavior of downloading.


Before I make changes in that direction: in my mind, a more flexible solution should do the following: upon clicking the new button, open a native file dialog box. There, the export file type will default to the one in `rcParams`, but the file type dropdown should allow for a variety of choices supported by `savefig`. 
It's clear to me this will involve one or two additional steps and a bit more coding, but probably worth it. Do you agree?

Issue #261, opened 2020-09-06
Double-click events are not recognised


ianhi:
Originally posted by @nvaytet in https://github.com/matplotlib/matplotlib/issues/18404

Posting here because ipympl will not automatically inherit whatever solution gets implemented for nbagg

> ### Bug report
> 
> **Bug summary**
> 
> Using the first example from the [events documentation](https://matplotlib.org/3.3.1/users/event_handling.html) detects double-click events properly in the Qt backend, but when used in a Jupyter notebook, the two clicks are recognised as two separate single click events.
> 
> **Code for reproduction**
> 
> I've adapted the code to show the contents of the `event` as the figure title instead of a `print` statement, because the output of `print` statements inside functions triggered by events disappear in Jupyter notebooks.
> 
> ```python
> import matplotlib.pyplot as plt
> import numpy as np
> fig, ax = plt.subplots()
> ax.plot(np.random.rand(10))
> 
> def onclick(event):
>     ax.set_title('%s click: button=%d, x=%d, y=%d, xdata=%f, ydata=%f' %
>           ('double' if event.dblclick else 'single', event.button,
>            event.x, event.y, event.xdata, event.ydata))
> 
> cid = fig.canvas.mpl_connect('button_press_event', onclick)
> ```
> 
> **Actual outcome**
> 
> The title always says `single-click`, no matter how fast I do my double-clicking.
> 
> ![Screenshot at 2020-09-04 08-02-20](https://user-images.githubusercontent.com/39047984/92204984-ffa02300-ee84-11ea-9f2d-c4a932cc91cc.png)
> 
> The double click events are properly captured when using the standard Qt window renderer.
> I have tried both the default `%matplotlib notebook` backend and the `ipympl` widget backend, and they show identical behaviour.
> 
> **Matplotlib version**
> 
>     * Operating system: Ubuntu 20.04
> 
>     * Matplotlib version: 3.2.2
> 
>     * Matplotlib backend (`print(matplotlib.get_backend())`): nbAgg, ipympl
> 
>     * Python version: 3.7.6
> 
>     * Jupyter version (if applicable): jupyter_client=6.1.6, jupyter_core=4.6.3, notebook=6.0.3
> 
>     * Other libraries:
> 
> 
> Installed in a `conda` env from `conda-forge`.

nvaytet:
Is there any update on this? It has been fixed in nbagg for a while.
Thanks!

ianhi:
Should be released as `0.9.0` now!

Issue #416, opened 2022-01-13
Incompatibilities between patch versions when installed in both server and kernel environment


wholtz:
This issue was originally discussed in [#34](https://github.com/matplotlib/ipympl/issues/34#issuecomment-1012369684_) but I now see this is a distinct issue.

If a jupyterlab user wants to make use of a custom kernel and use ipympl in their notebooks, then ipympl must be installed in both the jupyterlab environment and their kernel environment. If different versions of ipympl are installed in the jupyterlab environment and kernel environments, then sometimes ipympl breaks even for differences in the patch version. If jupyterlab is running within a multi-user environment (jupyterhub) and the user does not control the server, then the user may not know what version of ipympl is installed in the jupyterlab environment.  This can make it difficult for users to determine the version of ipympl they need to install in their kernels. Additionally, the kernel environments are likely to break when upgrades occur to ipympl in the jupyterlab environment.

## Test setup

The test were run using 6 builds of the following Dockerfile, where LAB_IPYMPL_VERSION modified for each build.

``` Dockerfile
# this is a micromamba:0.19.1 image, using digest instead of tag for reproducibility
FROM mambaorg/micromamba@sha256:1ab2666fcd1559e9865d599433671f97cb735103af9351e5e10a3f0d5b6048c8

# version of ipympl to install into the jupyterlab environment
ARG LAB_IPYMPL_VERSION=0.8.0

# Create conda environments:
# "kernelXXX" environments will be available in jupyter notebooks
# "lab" environment will run the jupyterlab server (created at line 56)

RUN micromamba create --yes --name kernel085 --channel conda-forge \
       ipympl=0.8.5 \
       ipywidgets=7.6.5 \
       matplotlib=3.5.0 \
       python=3.10.1 && \
    micromamba create --yes --name kernel084 --channel conda-forge \
       ipympl=0.8.4 \
       ipywidgets=7.6.5 \
       matplotlib=3.5.0 \
       python=3.10.1 && \
    micromamba create --yes --name kernel083 --channel conda-forge \
       ipympl=0.8.3 \
       ipywidgets=7.6.5 \
       matplotlib=3.5.0 \
       python=3.10.1 && \
    micromamba create --yes --name kernel082 --channel conda-forge \
       ipympl=0.8.2 \
       ipywidgets=7.6.5 \
       matplotlib=3.5.0 \
       python=3.10.1 && \
    micromamba create --yes --name kernel081 --channel conda-forge \
[...]

PR #446, opened 2022-03-31
support double click events

ianhi:
closes: #261  attn @nvaytet

@martinRenou for version numbers what do we do here? I think we need to bump package.json 
https://github.com/matplotlib/ipympl/blob/21a1c8dde81e5732606c41054285ab2def9dba9f/package.json#L3

to either `0.11.0` or `0.10.6` and also bump the python version. Is there anything we can do here to ameliorate the patch version incompatibility (#416)?

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
         "ts-jest": "^26.0.0",
         "ts-loader": "^8.0.0",
         "typescript": "~4.1.3",
-        "webpack": "^5.0.0",
+        "webpack": "^5.61.0",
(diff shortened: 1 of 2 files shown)


ianhi:
@nvaytet will release soon hopefully - going to try to work out whats happening with #416 first

martinRenou:
Sorry I missed your question. I will push a PR for that OpenSSL issue, we fixed it in multiple places.

nvaytet:
@ianhi awesome! thanks :-)

PR #447, opened 2022-03-31
update binder bot link

ianhi:


ianhi:
looks like it doesn't update until merged - so just gotta trust that it'll work

PR #448, opened 2022-03-31
Add model version separate from package version

### LINKED ISSUE #416: Incompatibilities between patch versions when installed in both server and kernel environment

ianhi:
Hopefully fixes #416 

1. Adds a model version version that is not tied to the npm release version so that we can (hopefully) have compatibility across patch versions.

2. Bumps the version for release of the double click fixes

### DIFF
--- a/ipympl/_version.py
+++ b/ipympl/_version.py
@@ -1,3 +1,9 @@
 version_info = (0, 8, 8)
 __version__ = '.'.join(map(str, version_info))
-js_semver = '^0.10.5'
+
+
+# The versions of protocol of communication with the frontend that this python verison knows
+# how to speak. See counterpart in the src/version.ts file.
+# These should not be changed unless we introduce changes to communication between
+# frontend and backend.
+__MODEL_VERSION__ = "1.0.0"
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -4,7 +4,7 @@ import { Widget } from '@phosphor/widgets';
 
 import { IJupyterWidgetRegistry } from '@jupyter-widgets/base';
 
-import { MODULE_NAME, MODULE_VERSION } from './version';
+import { MODEL_VERSION, MODULE_NAME } from './version';
 
 const EXTENSION_ID = 'matplotlib-jupyter:main';
 
@@ -31,7 +31,7 @@ function activateWidgetExtension(
 ): void {
     registry.registerWidget({
         name: MODULE_NAME,
-        version: MODULE_VERSION,
-        exports: () => import('./index'),
+        version: MODEL_VERSION,
+        exports: (): any => import('./index'),
     });
 }
--- a/src/toolbar_widget.ts
+++ b/src/toolbar_widget.ts
@@ -1,6 +1,6 @@
 import { DOMWidgetModel, DOMWidgetView } from '@jupyter-widgets/base';
 
-import { MODULE_VERSION } from './version';
+import { MODEL_VERSION } from './version';
 
 import '../css/mpl_widget.css';
 
@@ -12,8 +12,8 @@ export class ToolbarModel extends DOMWidgetModel {
             _view_name: 'ToolbarView',
             _model_module: 'jupyter-matplotlib',
             _view_module: 'jupyter-matplotlib',
-            _model_module_version: '^' + MODULE_VERSION,
-            _view_module_version: '^' + MODULE_VERSION,
+            _model_module_version: MODEL_VERSION,
+            _view_module_version: MODEL_VERSION,
             toolitems: [],
             position: 'left',
             button_style: '',
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
     hooks:
       - id: isort
   - repo: https://github.com/psf/black
-    rev: 21.12b0
+    rev: 22.3.0
     hooks:
       - id: black
   - repo: https://github.com/asottile/pyupgrade
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -58,7 +58,7 @@ from traitlets import (
     observe,
 )
 
-from ._version import js_semver
+from ._version import __MODEL_VERSION__
 
 cursors_str = {
     cursors.HAND: 'pointer',
@@ -93,11 +93,11 @@ def connection_info():
 class Toolbar(DOMWidget, NavigationToolbar2WebAgg):
 
     _model_module = Unicode('jupyter-matplotlib').tag(sync=True)
-    _model_module_version = Unicode(js_semver).tag(sync=True)
+    _model_module_version = Unicode(__MODEL_VERSION__).tag(sync=True)
     _model_name = Unicode('ToolbarModel').tag(sync=True)
 
     _view_module = Unicode('jupyter-matplotlib').tag(sync=True)
-    _view_module_version = Unicode(js_semver).tag(sync=True)
+    _view_module_version = Unicode(__MODEL_VERSION__).tag(sync=True)
     _view_name = Unicode('ToolbarView').tag(sync=True)
 
     toolitems = List().tag(sync=True)
@@ -180,11 +180,11 @@ class Toolbar(DOMWidget, NavigationToolbar2WebAgg):
 class Canvas(DOMWidget, FigureCanvasWebAggCore):
 
     _model_module = Unicode('jupyter-matplotlib').tag(sync=True)
-    _model_module_version = Unicode(js_semver).tag(sync=True)
+    _model_module_version = Unicode(__MODEL_VERSION__).tag(sync=True)
     _model_name = Unicode('MPLCanvasModel').tag(sync=True)
 
     _view_module = Unicode('jupyter-matplotlib').tag(sync=True)
-    _view_module_version = Unicode(js_semver).tag(sync=True)
+    _view_module_version = Unicode(__MODEL_VERSION__).tag(sync=True)
     _view_name = Unicode('MPLCanvasView').tag(sync=True)
 
     toolbar = Instance(Toolbar, allow_none=True).tag(sync=True, **widget_serialization)
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ setup_args = dict(
         'pillow',
         'traitlets<6',
         'ipywidgets>=7.6.0,<8',
-        'matplotlib>=2.0.0,<4',
+        'matplotlib>=3.4.0,<4',
     ],
     extras_require={
         "docs": [
--- a/src/mpl_widget.ts
+++ b/src/mpl_widget.ts
@@ -10,7 +10,7 @@ import {
 
 import * as utils from './utils';
 
-import { MODULE_VERSION } from './version';
+import { MODEL_VERSION } from './version';
 
 import { ToolbarView } from './toolbar_widget';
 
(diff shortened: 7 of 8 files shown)


ianhi:
An alternative approach that would make things simpler going forward would be to declare the current state as being model version `1.0.0` so we have the full range of semver to play with when making changes. The downside would be that we wouldn't get the back compatibility for this release. Not sure the best tradeoff there - curious for thoughts from @martinRenou and @wholtz

ipympl/_version.py
```diff
@@ -1,3 +1,10 @@
...
 __version__ = '.'.join(map(str, version_info))
-js_semver = '^0.10.5'
+
+
+# The versions of protocol of communication with the frontend that this python verison knows
+# how to speak. See counterpart in the src/version.ts file.
+# These should not be changed unless we introduce changes to communication between
+# frontend and backend.
+__TOOLBAR_MODEL_VERSION__ = "^0.10.5"
+__CANVAS_MODEL_VERSION__ = "^0.10.5"
```
martinRenou, line 10: It might be easier to track if we have one same version for all models?

README.md
```diff
@@ -64,6 +64,7 @@ Versions lookup table:
 
 | `ipympl` | `jupyter-matplotlib` | `JupyterLab` | `Matplotlib` |
 |----------|----------------------|--------------|--------------|
+| 0.8.9    | 0.10.x               | 3 or 2       | 3.4.0>=      |
```
martinRenou, line 67: I would prefer if we do version bumps separately, if you don't mind

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "jupyter-matplotlib",
-    "version": "0.10.5",
+    "version": "0.10.6",
```
martinRenou, line 3: Same comment here for the version bump

martinRenou (commented):
Thanks a lot for doing this!

martinRenou:
Concerning the version bump, what I did until now was to push release commits to master ([see this for example](https://github.com/matplotlib/ipympl/commit/5abd20ff41faa664e37a6a0452bf20ef66908d10)), that keeps the release commit clean in the history and far from any merge commit.

README.md
ianhi: :+1:

ipympl/_version.py
ianhi: Yeah that's probably a good call. One isn't really functional without the other so no need to have them separate.

ianhi:
- removed version bumping stuff
- Used same number for toolbar and canvas
- model version now `1.0.0`

ianhi:
any idea what's causing this new error?

```
src/plugin.ts:35:9 - error TS2322: Type '() => Promise<typeof import("/home/ian/Documents/oss/matplotlib/ipympl/src/index")>' is not assignable to type 'ExportData'.
  Type '() => Promise<typeof import("/home/ian/Documents/oss/matplotlib/ipympl/src/index")>' is not assignable to type '() => ExportMap'.
    Type 'Promise<typeof import("/home/ian/Documents/oss/matplotlib/ipympl/src/index")>' is not assignable to type 'ExportMap'.
      Index signature is missing in type 'Promise<typeof import("/home/ian/Documents/oss/matplotlib/ipympl/src/index")>'.

35         exports: () => import('./index'),
           ~~~~~~~

  node_modules/@jupyter-widgets/base/lib/registry.d.ts:35:5
    35     exports: ExportData;
           ~~~~~~~
    The expected type comes from property 'exports' which is declared here on type 'IWidgetRegistryData'


```

src/plugin.ts
```diff
@@ -32,6 +32,6 @@ function activateWidgetExtension(
     registry.registerWidget({
         name: MODULE_NAME,
         version: MODULE_VERSION,
-        exports: () => import('./index'),
+        exports: (): any => import('./index'),
```
ianhi, line 35: this gets rid of the typescript error, but is probably not the best way to do that.

ianhi:
Thank goodness for those galata tests. they catch that this actually does break things :/

ipympl/_version.py
```diff
@@ -1,3 +1,9 @@
 version_info = (0, 8, 8)
 __version__ = '.'.join(map(str, version_info))
-js_semver = '^0.10.5'
+
+
+# The versions of protocol of communication with the frontend that this python verison knows
+# how to speak. See counterpart in the src/version.ts file.
+# These should not be changed unless we introduce changes to communication between
+# frontend and backend.
+__MODEL_VERSION__ = "^1.0.0"
```
ianhi, line 9: in ipywidgets there is no `^` maybe we should remove also?

martinRenou:
> I suspect there isn't a way through this change without a bit of pain for users and I'd lean towards getting the painful parts over soon

I actually believe this could go smoothly for users. They will hopefully just see a version bump on ipympl and jupyter-matplotlib, as for any release. Then it will hopefully be easier for them to update one without the other from now on :)

src/toolbar_widget.ts
```diff
@@ -12,8 +12,8 @@ export class ToolbarModel extends DOMWidgetModel {
             _view_name: 'ToolbarView',
             _model_module: 'jupyter-matplotlib',
             _view_module: 'jupyter-matplotlib',
-            _model_module_version: '^' + MODULE_VERSION,
-            _view_module_version: '^' + MODULE_VERSION,
+            _model_module_version: MODEL_VERSION,
```
ianhi, line 15: I'm not sure having the `^` does anything - and ipywidgets doesn't do this, so I removed it to be more similar to ipywidgets.

PR #450, opened 2022-04-01
Fix OpenSSL issue

martinRenou:
cc. @ianhi


ianhi:
Is this a better fix than https://github.com/matplotlib/ipympl/commit/1e110c9b9cf5f5aefc94a5694a8c652a4af65699 ? If so should we revert that webpack requirement?

PR #466, opened 2022-05-22
Remove unused 'close' arg to show().

anntzer:
For reference, _Backend_ipympl.show is effectively the function that
pyplot.show calls (pyplot.show even copies the signature of the backend
show).

In af699c2, a 'close' kwarg was added to _Backend_ipympl.show, although
any use of it was immediately removed in a subsequent commit (783419b).
Still, this change means that calling e.g. `plt.show(True)`, for
example, would no longer set `block=True`, but instead set `close=True`.
As the change is relatively recent, it seems better to remove the unused
'close' kwarg again, which also makes the signature consistent with the
builtin Matplotlib backends.

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -498,7 +498,7 @@ class _Backend_ipympl(_Backend):
         return manager
 
     @staticmethod
-    def show(close=None, block=None):
+    def show(block=None):
         # # TODO: something to do when keyword block==False ?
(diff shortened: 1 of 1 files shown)


ianhi:
Makes sense to me. Thanks for updating here as you go on the main repo :)

Only concern is the visual regression test failing. Maybe  arestart will help

ianhi (approved):
ah indeed good catch! thanks again!

PR #468, opened 2022-06-02
Move draw_if_interactive logic to new_figure_manager_given_figure.

anntzer:
Currently, Matplotlib only ever calls draw_if_interactive at the end of
`plt.figure()` (and upon figure unpickling), i.e. this is a mechanism to
further customize handling of newly generated figures.  (Prior to
Matplotlib 1.5, draw_if_interactive was also called at the end of all
pyplot functions to trigger a figure redraw, but this is now handled by
the stale attribute.)

In order to simplify the backend API ("what is the API that a backend
module must/can provide", I am planning to deprecate (on Matplotlib's
side) the ability for backends to provide a draw_if_interactive
function (forcing them to always do `if interactive(): draw_idle()`) (https://github.com/matplotlib/matplotlib/issues/23105).
Instead, any relevant new-figure-customization logic can instead go
into `new_figure_manager_given_figure`.  This PR implements this change
for ipympl, and should be fully back-compatible all the way back to
Matplotlib 1.5.  I would like to make these changes first on the side
of the clients (i.e., the third-party backends), to catch any possible
problems with the intended change on Matplotlib's side.

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -495,6 +495,21 @@ class _Backend_ipympl(_Backend):
             Gcf.destroy(manager)
 
         cid = canvas.mpl_connect('close_event', destroy)
+
+        # Only register figure for showing when in interactive mode (otherwise
+        # we'll generate duplicate plots, since a user who set ioff() manually
+        # expects to make separate draw/show calls).
+        if is_interactive():
+            # ensure current figure will be drawn.
+            try:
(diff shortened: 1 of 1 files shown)


anntzer:
@ianhi any chance you can have a look at this? :)

ianhi (approved):
seems good to me! thakns for the ping

ianhi:
and thanks also for the PR! makes following the backend changes in matplotlib much nicer :)

PR #478, opened 2022-07-16
Ensure throttling does not break `capture_scroll`

ianhi:
Since we started throttling mouse events we now also throttle the `capture_scroll` This PR adds that as a separate `EventListener` that is not throttled.


martinRenou (approved):
LGTM! Thanks!

PR #494, opened 2022-11-15
MNT: PEP 632 removed distutils

tacaswell:
It was only being used for logging.


martinRenou (approved):
Thank you! These logs were bringing issues when installing the labextension in develop mode anyway, so we should remove them.

martinRenou:
CI failures are unrelated. It looks like we are pulling Python 3.11.

PR #505, opened 2022-12-19
Make handle_binary compliant with the KernelMessage.IMessage spec

mctoohey:
I found that `ipympl` was failing to display plots on more recent versions of the Jupyterlab 4.0 prereleases. This pull request fixes an issue where the  `handle_binary` expects a `DataView` but new versions of Jupyterlab will pass it an `ArrayBuffer`.

Since Jupyterlab v4.0.0a21 the buffers deserialized from a `KernelMessage` are now of type `ArrayBuffer` instead of the `DataView` type `handle_binary` expects. However, it turns out that the spec for `KernelMessage.IMessage` has always specified that buffers could be either an `ArrayBuffer` or an `ArrayBufferView`. This commit ensures either type is accepted.

### DIFF
--- a/src/mpl_widget.ts
+++ b/src/mpl_widget.ts
@@ -234,10 +234,10 @@ export class MPLCanvasModel extends DOMWidgetModel {
         this.send_draw_message();
     }
 
-    handle_binary(msg: any, dataviews: any) {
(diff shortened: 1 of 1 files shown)


martinRenou (approved):
Thanks! The CI failures are unrelated (Python 3.11 broke our CI, we need to fix it separately)

PR #514, opened 2023-02-07
Fixup CI

martinRenou:


martinRenou:
update galata references

ianhi:
is it possible to easily look at the differences in the generated images? I suspect that were catching a styling change in matplotlib latest

martinRenou:
You can download the test artifacts on this page https://github.com/matplotlib/ipympl/actions/runs/4122954626 (bottom of the page -> ui-test-output file). I will contain the expected image, the actual image and the diff.

There seems to be a slight difference in the font rendering of the axis labels.

> I suspect that were catching a styling change in matplotlib latest

Yes, I was thinking the same. It seems to only happen on Ubuntu.

Because it's a very slight difference I tried to change the threshold for the diff, without success.

ianhi:
> Yes, I was thinking the same. It seems to only happen on Ubuntu.

I think that there have been a bunch of changes to the font system recently. I almost wonder if this is surfacing a very subtle bug upstream

ianhi:
@martinRenou what do you think about merging this now and fixing up the ubuntu latest in the future? As it stands this allows us to run tests, which we otherwise cant. which would be helpful for https://github.com/matplotlib/ipympl/pull/517 and https://github.com/matplotlib/ipympl/pull/478

martinRenou:
That's fine with me, half working CI is better than non working one

PR #520, opened 2023-02-23
add bot to comment when cron tests fail

ianhi:


ianhi:
cant really test until we merge - so here we go

PR #521, opened 2023-03-01
Minor doc formatting improvement

oscargus:
While reviewing https://github.com/matplotlib/matplotlib/pull/25353 I realized that it would look better if both installation options were formatted the same.


ianthomas23 (approved):
Thanks for this @oscargus. Sorry it wasn't looked at any quicker.

PR #528, opened 2023-06-15
Fix typo in "full-example.ipynb"

yvonnefroehlich:
This PR aims to fix a typo in the Jupyter notebook `full-example.ipynb`.


ianhi:
nice catch thanks for the fix!

PR #541, opened 2024-01-23
Update to hatch-based build system and fix CI

ianthomas23:
This PR switches the build system from `jupyter_packaging` to the recommended `hatch_jupyter_builder`, plus some other smaller changes, in an attempt to get CI passing. When it succeeds and this PR is merged, I will deal with other improvements such as bringing dependencies up to date.

Summary of changes made:
- Switch to `hatch`-based builder.
- Update `micromamba` github actions.
- Use `python -m build` to build the package rather than the deprecated `python setup.py ...`
- Use `jlpm` rather than `yarn` for TypeScript package management and building.

With these changes I can build and run the tests locally, and also install from the generated sdist and binary wheel.


pyproject.toml
```diff
@@ -1,7 +1,98 @@
...
+    "Framework :: IPython",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Science/Research",
+    "Programming Language :: Python :: 3.5",
+    "Programming Language :: Python :: 3.6",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
martinRenou, line 33: Should we add Python 3.12 here?

ianthomas23: Yes, I don't know why I stopped at 3.11.

We should probably decide on what Python versions we are supporting, heading toward [SPEC0](https://scientific-python.org/specs/spec-0000/) eventually. Maybe 3.9 to 3.12 is appropriate now. Then I could ensure that CI is running using both the earliest and latest Python versions that we claim to support.

ianthomas23:
> Would be nice to setup the jupyter-releaser after that

I'll take a look at jupyter-releaser.

pyproject.toml
martinRenou: Sounds good 👍🏽

martinRenou:
Bot update galata references

martinRenou:
The bot is failing for some other reason. Anyway I believe the CI failures are unrelated. It seems the visual regression diff is super small and probably due to some update on the matplotlib side. I can look into getting this green separately.

ianthomas23:
I am going to look at excluding text from the Matplotlib plots. This is done in Matplotlib CI as there can be tiny differences depending on platform, OS and compiler.

martinRenou:
Thank you! I guess this PR is good to go as-is!

> Fix update_galata_references.yaml

The fix could probably be just https://github.com/jupytercad/jupytercad/blob/main/.github/workflows/update_galata_references.yaml#L31

.github/workflows/main.yml
```diff
@@ -49,25 +46,28 @@ jobs:
...
         include:
-          - python: '3.10'
-            dist: 'ipympl*.tar.gz'
+          # Earliest supported Matplotlib, needs python <= 3.10
           - python: '3.9'
+            mpl-version: '3.5'
             dist: 'ipympl*.whl'
+          # Install from sdist not wheel
+          - python: '3.12'
+            mpl-version: 'latest'
```
martinRenou, line 59: Are we missing an `os` entry here? I wonder if this is the reason why the workflow does not start

ianthomas23: Yes, I've corrected this now.

Could you cancel the old CI runs as I don't have permissions to do that?

ianthomas23:
I changed the test that installed from sdist from python 3.12 to 3.11 so now it passes. With python 3.12 it was trying to compile zmq from source. I don't think that should be necessary, but it is not a problem for us to solve today.

.github/workflows/main.yml
ianthomas23: Thanks!

martinRenou:
> I changed the test that installed from sdist from python 3.12 to 3.11 so now it passes. With python 3.12 it was trying to compile zmq from source. I don't think that should be necessary, but it is not a problem for us to solve today.

We may be able to workaround this issue by installing from source with `--no-deps` and making sure all dependencies are installed using conda/mamba from conda-forge.

Thank you for getting this green! Merging and we can iterate

martinRenou (approved):
Thank you!

PR #543, opened 2024-01-26
Build using JupyterLab 4

### LINKED ISSUE #539: Please remove jupyterlab==3.* requirement so as to allow 4.x to work.
This is basically a duplicate of https://github.com/matplotlib/ipympl/issues/532 but I don't think the original intent was understood (and I can't reopen that one as I'm not affiliated). It appears ipympl works with jupyterlab if you install it without checking deps (which can be extremely frustrating for certain tools). In pyproject.toml you have:

requires = ["jupyter_packaging~=0.7.0", "jupyterlab>=3.0.0,==3.*", "setuptools>=40.8.0", "wheel"]

Changing this to:

requires = ["jupyter_packaging~=0.7.0", "jupyterlab>=3.0.0", "setuptools>=40.8.0", "wheel"]

Would be greatly appreciated!

ianthomas23:
Closes #539.

This changes the build system to build against JupyterLab 4 rather than 3. I have manually checked that the wheel generated in this way works on all the following:

- JupyterLab 4
- JupyterLab 3
- Notebook 7
- Notebook 6
- nbclassic

Note that `ipympl` already worked fine on JupyterLab 4, this just modernises the build process.


ianthomas23:
I've removed the captions above the Matplotlib figures otherwise slight differences in rendering between ubuntu and macos cause the image comparison tests to fail. These captions are in `<div>`s anyway rather than generated by Matplotlib.

martinRenou (approved):
Thank you!

PR #549, opened 2024-03-12
Add matplotlib.backend entry points

ianthomas23:
This adds entry points to `pyproject.toml` for ipympl to register itself as a Matplotlib backend using the names  "widget" and "ipympl". It is part of the planned transition of backend mapping from IPython to Matplotlib (https://github.com/matplotlib/matplotlib/issues/27663 and https://github.com/ipython/ipython/issues/14311).

I have bumped the package __version__ as it will be used by Matplotlib to determine if a backward-compatible shim is required for this backend. I have also bumped some of the `pre-commit` hook versions which was required for `flake` to be happy on my dev machine.

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -540,7 +540,8 @@ class _Backend_ipympl(_Backend):
 
 
 def flush_figures():
-    if rcParams['backend'] == 'module://ipympl.backend_nbagg':
+    backend = matplotlib.get_backend()
+    if backend in ('widget', 'ipympl', 'module://ipympl.backend_nbagg'):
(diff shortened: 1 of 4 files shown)


ipympl/backend_nbagg.py
```diff
@@ -540,7 +540,8 @@ def show(block=None):
 
 
 def flush_figures():
-    if rcParams['backend'] == 'module://ipympl.backend_nbagg':
+    backend = matplotlib.get_backend()
```
ianthomas23, line 543: This is considered the correct way to determine the current Matplotlib backend, it just calls `rcParams['backend']` internally.

PR #550, opened 2024-03-15
Use correct versioning

ianthomas23:
When I updated the build system to use `hatch` (#541) I didn't set up the dynamic versioning correctly. This came to light in #549 as currently the package version is `0.9.4.dev1` but `ipympl.__version__` is still `0.9.3`.

This fixes the dynamic versioning in `pyproject.toml` so that `hatch` reads the one-and-only version number from `ipympl._version.py`.


martinRenou (approved):
Thanks!

PR #555, opened 2024-05-28
Update binder2

ianthomas23:
Update the binder config, both in the top of the README and in the github actions PR workflow. Take 2.

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -6,3 +6,4 @@ dependencies:
   - jupyterlab~=4.0
   - ipywidgets>=7.6
   - matplotlib
+  - notebook>=7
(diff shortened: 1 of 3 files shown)


ianthomas23:
Here is the mybinder link that the README will effectively point to (but it will be on `main` rather than this PR branch): https://mybinder.org/v2/gh/ianthomas23/ipympl/update-binder2?urlpath=lab/tree/docs/examples/full-example.ipynb. It all seems to work OK.

It appears that the binder workflow is using the github action from `main` rather than this branch so it won't work until this is merged. I am happy to merge and iterate.

martinRenou:
Thanks for looking into it!

PR #558, opened 2024-07-02
Remove unnecessary dependency `ipython_genutils`

randolf-scholz:
I noticed this project depends on `ipython_genutils`, which was last updated in 2017.
Moreover, the only used thing from `ipython_genutils` was `ipython_genutils.py3compat.string_types`, which happens to be the tuple `(str,)`.

```python
from ipython_genutils.py3compat import string_types
assert string_types == (str,)  # ✅
assert string_types[0] is str  # ✅
```

Also removed a redundant python 2 check.


martinRenou (approved):
Thanks!

PR #569, opened 2024-12-17
Update ui-test dependencies

ianthomas23:
`ui-tests` are currently failing, initially not able to `page.filebrowser.openDirectory()`. I think this is due to old `galata` test dependencies not working with latest JupyterLab that is used to build and run. Updating the test dependencies works for me locally, so trying here in CI.


martinRenou (approved):
Thanks!

PR #571, opened 2024-12-17
Support python 3.13 in classifiers and CI

ianthomas23:
`ipympl` is working fine for me locally using python 3.13, so here adding a CI run for it on both `macos` and `ubuntu`, and updating the classifiers.


martinRenou (approved):
Thanks!

PR #577, opened 2025-01-02
Remove workaround for matplotlib < 3.4

### LINKED ISSUE #575: key_press_event not working with matplotlib 3.10
## Describe the issue
fig.canvas.mpl_connect('key_press_event', ...) does not work anymore with matplotlib 3.10. I compiled the following example to demonstrate this (run in a jupyter notebook):
```
%matplotlib widget
import matplotlib

from ipywidgets import Output
from IPython.display import display

plt.close('test')
fig,ax = plt.subplots(num='test')

output = Output()
def event_handler(event):
    with output:
        print(event)

fig.canvas.mpl_connect('key_press_event', event_handler)
fig.canvas.mpl_connect('button_release_event', event_handler)

display(output)
```

When using matplotlib 3.9 the key press event and the mouse button release event cause the event_handler to print the event to the output. After upgrading to matplotlib 3.10, the button_release_event continues to work, but the key_press event does not produce any output anymore. When changing the matplotlib backend to qt (linemagic: "%matplotlib qt") the events work both for matplotlib 3.9 and 3.10

## Versions

```
 3.12.8 | packaged by conda-forge | (main, Dec  5 2024, 14:24:40) [GCC 13.3.0]
ipympl version: 0.9.5
Selected Jupyter core packages...
IPython          : 8.30.0
ipykernel        : 6.29.5
ipywidgets       : 8.1.5
jupyter_client   : 8.6.3
jupyter_core     : 5.7.2
jupyter_server   : 2.14.2
jupyterlab       : 4.3.4
nbclient         : 0.6.8
nbconvert        : 7.16.4
nbformat         : 5.10.4
notebook         : not installed
traitlets        : 5.14.3
Known nbextensions:
  config dir: /etc/jupyter/nbconfig
    notebook section
      - Validating: problems found:
        - require?  X jupyter-js-widgets/extension
JupyterLab v4.3.4
/ssd/opt/miniforge3/envs/dev/share/jupyter/labextensions
        jupyter-matplotlib v0.11.5 enabled OK
        jupyterlab_pygments v0.3.0 enabled OK (python, jupyterlab_pygments)
        jupyter-leaflet v0.19.2 enabled OK
        jupyterlab-execute-time v3.2.0 enabled OK (python, jupyterlab_execute_time)
        @jupyter-widgets/jupyterlab-manager v5.0.13 enabled OK (python, jupyterlab_widgets)

```

ianthomas23:
Fixes #575.

Workaround code for matplotlib < 3.4 was using a string check of version number:

https://github.com/matplotlib/ipympl/blob/efd2e5cdd0e814989fdeb79a3fcb3744ff2c1d50/ipympl/backend_nbagg.py#L387

which breaks using matplotlib 3.10.

It is possible to replace this with 
```python
if matplotlib.__version_info__ < (3, 4):
```
but `__version_info__` was introduced into matplotlib at version 3.5. Anyway, we no longer support matplotlib < 3.4

https://github.com/matplotlib/ipympl/blob/efd2e5cdd0e814989fdeb79a3fcb3744ff2c1d50/pyproject.toml#L47

so I have just removed the workaround in this PR, which fixes the problem locally for me.


ipympl/backend_nbagg.py
```diff
@@ -384,71 +384,6 @@ def _ipython_display_(self, **kwargs):
...
-            if e_type == 'key_press':
-                self.key_press_event(key, guiEvent=guiEvent)
-            elif e_type == 'key_release':
-                self.key_release_event(key, guiEvent=guiEvent)
-
-        handle_key_press = handle_key_release = _handle_key
-
 
 class FigureManager(FigureManagerWebAgg):
     if matplotlib.__version__ < "3.6":
```
jenshnielsen, line 389: Does this line not have the same issue?

ianthomas23: Ah yes, I missed that. It doesn't seem to cause a problem though, in my manual testing.

#580 fixes it.

Issue #488, opened 2022-09-12
Warnings with backend_nbagg


lrdossan:
## Issue

There is a warning being triggered when running `pytest` with the following instruction:

```
ipympl.backend_nbagg.new_figure_manager_given_figure(1, matplotlib.figure.Figure()))
```

The warning is the following:

```
 /Users/luisaleixo/.pyenv/versions/3.9.6/envs/cara/lib/python3.9/site-packages/ipywidgets/widgets/widget.py:477: DeprecationWarning: Passing unrecognized arguments to super(Toolbar).__init__().
  __init__() missing 1 required positional argument: 'canvas'
  This is deprecated in traitlets 4.2.This error will be raised in a future release of traitlets.
    super(Widget, self).__init__(**kwargs)
```

It is not possible to downgrade the triatlets version for this project.

## Versions

Name: ipympl
Version: 0.9.1

Name: pytest
Version: 6.2.5

miracleyoo:
Same issue here. No matter what I plot in vscode-opened notebook using `%matplotlib widget`, this error msg pops out.

PR #595, opened 2025-04-30
Fix: missing 1 required positional argument: 'canvas' in Toolbar init

fleming79:
Resolves: https://github.com/matplotlib/ipympl/issues/488


fleming79:
@ianthomas23  - thank you for running the workflow. Unfortunately, I don't think the patched version was used in the test even though it was downloaded. I've changed it to force-install the patch just prior to running the test in the workflow. Could you please trigger the workflow again?

fleming79:
Thanks @ianthomas23  - The patched version of Ipywidgets requires Python >= 3.10 which explains the failing tests.

PR #600, opened 2025-07-13
bug_report.md: Use set -x and use `;` instead of `&&`

pcworld:
set -x shows the command in the output, making the output's respective command clearer. Using `;` rather than `&&` also runs the other commands even if one fails, for example if "jupyter-nbextension" is not found.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -17,7 +17,7 @@ In general, please make sure you are using the latest versions of software when 
 Please include the result of the following command, which will print out the version of python, ipympl and the Jupyter ecosystem.
 
 ```bash
-python -c "import sys; print('\n',sys.version); import ipympl; print('ipympl version:', ipympl.__version__)" && jupyter --version && jupyter nbextension list && jupyter labextension list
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -17,7 +17,7 @@ In general, please make sure you are using the latest versions of software when
 Please include the result of the following command, which will print out the version of python, ipympl and the Jupyter ecosystem.
 
 ```bash
-python -c "import sys; print('\n',sys.version); import ipympl; print('ipympl version:', ipympl.__version__)" && jupyter --version && jupyter nbextension list && jupyter labextension list
+set -x ; python -c "import sys; print('\n',sys.version); import ipympl; print('ipympl version:', ipympl.__version__)" ; jupyter --version ; jupyter nbextension list ; jupyter labextension list
```
ianhi, line 20: This `set -x` seems to dramatically break the output for me on zsh

we need this command to be as shell agnostic as possible.

An alternative is to have a "print_debug_info" in ipympl so we could have `python -c "import ipympl; ipympl.print_debug_info()`


At minimum we need to remove set -x or find a more general way to get the same effect.

ianhi (changes requested):
Thanks very much for the PR! Unfortunately on my terminal (zsh) it broke some things so we need to find a way to deal with that

PR #603, opened 2025-08-13
Use thread lock to support comms via subshells

ianthomas23:
The upcoming `ipykernel` 7.0.0 release will support [subshells](https://jupyter.org/enhancement-proposals/91-kernel-subshells/kernel-subshells.html) which are separate threads of execution within a kernel process. JupyterLab has already been updated to deal with subshells on kernels that support them and this includes options for running comms over subshells (jupyterlab/jupyterlab#17363). This PR contains changes that are needed in `ipympl` to support subshells, specifically use of a `threading.Lock` in the python code that accesses global state such at Matplotlib's `Gcf` and `ipython.display`. Without this we see problems such as plots not being displayed as one thread can be, for example, modifying a collection in `Gcf` whilst another is reading it. Strictly speaking we only need to lock when one thread is writing to a global as we can tolerate multiple non-modifying reads at the same time, but I have preferred to keep the changes as simple and understandable as possible but putting the locks at a high level in each of the functions that needs it.

Screenshot 2025-08-13 at 11 29 40

Above is a screenshot reproducing the problem using latest commit 5e068ed, showing that the third plot is not displayed. To reproduce this use the latest `jupyterlab` and precisely `ipykernel==7.0.0a2`, and in Lab's Settings Editor select "Kernel comms over subshells" to be "One subshell per comm-target". It does not occur every time, but this is the setting that I have found most likely to produce the problem. After this PR, all plots display correctly.

### DIFF
--- a/ipympl/backend_nbagg.py
+++ b/ipympl/backend_nbagg.py
@@ -19,6 +19,7 @@ except ModuleNotFoundError:
 import io
 import json
 from base64 import b64encode
+from threading import Lock
 
 try:
     from collections.abc import Iterable
@@ -67,6 +68,10 @@ cursors_str = {
     cursors.WAIT: 'wait',
 }
 
+# threading.Lock to prevent multiple threads from accessing globals such as Gcf
+_lock = Lock()
+
+
 
 def connection_info():
     """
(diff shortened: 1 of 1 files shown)


ianthomas23:
We will need a new release of `ipympl` with this before `ipykernel 7.0.0` is released. I think it could be `0.9.8` as this isn't really new functionality, or `0.10.0` is that is preferred.

PR #613, opened 2026-01-14
Add Pixi setup

ianhi:
adds a pixi setup to make it easier to get a working dev env.

I also find that pixi plays much more nicely with coding agents than conda/mamba

I didn't add the lock file as I think it's mostly bloat for the development pattern of this repo. I'd also like to look into removing yarn.lock as a bloat (hasn't been udpated in 2 years)

🤖 changes are by claude fully directed by me


martinRenou (approved):
Thanks! No strong opinion on this myself. As long as we don't force devs to use something I'm good 👍🏽

PR #614, opened 2026-01-14
Update snapshots bot checkout and permission check

martinRenou:
Context in https://github.com/jupyterlab/maintainer-tools/pull/268


ianhi (approved):
seems like a ncie simplification

Issue #339, opened 2021-08-17
Download figure as pdf, not as png.


srosenzweig09:
## Describe the issue
Widgets are displayed with a toolbar, which has a save/download button that saves the figure directly to my Downloads folder. I love this feature, but I would appreciate it even more if I can set it to download the image as a pdf instead of as a png. Is this a possibility?

martinRenou:
I believe this is not on the roadmap for ipympl, but maybe someone has another opinion, do you have any input on this @ianhi? 

I guess this would be difficult to implement unless we find a JavaScript library that makes this easy.

martinRenou:
https://github.com/MrRio/jsPDF

ianhi:
This is the same underlying issue as https://github.com/matplotlib/ipympl/issues/138 and https://github.com/matplotlib/ipympl/issues/234 which is the savefig button doesn't respect rcparams because it doesn't follow the normal savefig path.

Ultimately I think ipympl should respect that from buttons. I think the way to have that happen is to have a bit more back and forth between front and backend when the save button is pushed. So it could like what I sketched out in https://github.com/matplotlib/ipympl/issues/138#issuecomment-663699886

ianhi:
For pdf we need the option 1 i proposed which is just using the javascript to choose the filepath and then doing everything on hte backend

abast:
Is there any news regarding this feature? pdf is the format I rely on, as I can import it as vectorgraphic into illustrator, and this worflow would be much improved if I could download a pdf directly from the notebook.

giulioiannelli:
I also found myself in the need of this functionality. Any news?

bjudkewitz:
+1

Issue #234, opened 2020-06-12
Saved figure backgrounds are always transparent


slopedog:
## Describe the issue

I'm not positive that this is ipympl and not matplotlib, but when I click the button to save the plot, the image always has a transparent background. Using `fig.patch.set_facecolor('#111111')` doesn't seem to help.

![image](https://user-images.githubusercontent.com/8258310/84513842-0d844480-ac98-11ea-883b-c31592154597.png)

## Versions

```
 3.6.8 (default, Apr 25 2019, 21:02:35) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]
ipympl version: 0.5.6
jupyter core     : 4.6.3
jupyter-notebook : 6.0.3
qtconsole        : 4.7.4
ipython          : 7.15.0
ipykernel        : 5.3.0
jupyter client   : 6.1.3
jupyter lab      : 2.1.4
nbconvert        : 5.6.1
ipywidgets       : 7.5.1
nbformat         : 5.0.7
traitlets        : 4.3.3
Known nbextensions:
  config dir: /home/jah/env/etc/jupyter/nbconfig
    notebook section
      jupyter-matplotlib/extension  enabled 
      - Validating: OK
      jupyter-js-widgets/extension  enabled 
      - Validating: OK
JupyterLab v2.1.4
Known labextensions:
   app dir: /home/jah/env/share/jupyter/lab
        @jupyter-widgets/jupyterlab-manager v2.0.0  enabled  OK
        jupyter-matplotlib v0.7.2  enabled  OK

```

ianhi:
I think this is the correct behavior, in the rcparams there is a separate setting for the facecolor of savefig. See https://stackoverflow.com/a/4805178/835607

slopedog:
That didn't seem to do the trick.
![image](https://user-images.githubusercontent.com/8258310/88406445-b301fc00-cd9e-11ea-86e4-bddbf8a52614.png)

ianhi:
Maybe this is a matplotlib issue? I am using matplotlib version 3.2.1  and get this:
![image](https://user-images.githubusercontent.com/10111092/88426747-b5288280-cdbf-11ea-8e23-b91dd341d341.png)

ianhi:
another check to isolate the cause is to try this with `%matplotlib inline` to see if this is due to ipympl or something else

slopedog:
[...]

PR #615, opened 2026-01-14
FIX: Toolbar savefig respects rcparams

ianhi:
Make the toolbar savefig button respect the matplotlib rcparams

Closes: https://github.com/matplotlib/ipympl/issues/138

and thereby 

Closes: https://github.com/matplotlib/ipympl/issues/234
Closes: https://github.com/matplotlib/ipympl/issues/339

## Implementation

Let the backend handle the actual buffer generation, then just add the correct mimetype to the bundle for download. Much simpler than i expected.

## Tests

The most important test is the manual_test notebook i added. I ran through this and it worked for me as expected. I will try it again with the PR binder link

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,8 @@ lib/
 ipympl/nbextension/index.*
 ipympl/nbextension/package.json
 ipympl/labextension/*.tgz
+.yarn/
+package-lock.json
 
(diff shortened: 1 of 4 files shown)


martinRenou (approved):
That's great! Thanks for this

